Issue #36, opened 2015-01-23
Getting optimisation issue with some CSS


MoOx:
When I use `webpack -p`, I'm getting a non helping error message

```
ERROR in ./~/css-loader!./~/cssnext-loader!./src/index.css
Module build failed: Error: Please check the validity of the CSS block starting from the line #1
    at throwError (/.../node_modules/css-loader/node_modules/csso/lib/gonzales.cssp.node.js:399:15)
    at checkStylesheet (/.../node_modules/css-loader/node_modules/csso/lib/gonzales.cssp.node.js:1921:22)
    at Object.CSSPRules.stylesheet (/.../node_modules/css-loader/node_modules/csso/lib/gonzales.cssp.node.js:365:40)
    at _getAST (/.../node_modules/css-loader/node_modules/csso/lib/gonzales.cssp.node.js:409:38)
    at exports.srcToCSSP (/.../node_modules/css-loader/node_modules/csso/lib/gonzales.cssp.node.js:2287:16)
    at Object.exports.srcToCSSP (/.../node_modules/css-loader/node_modules/csso/lib/gonzales.cssp.node.js:2292:16)
    at Object.exports.parse (/.../node_modules/css-loader/node_modules/csso/lib/cssoapi.js:7:21)
    at Object.module.exports (/.../node_modules/css-loader/index.js:18:18)
 @ ./src/index.css 4:14-293
```

In this index.css I just have some inlined `@import`, so this error is not really helping.
Is there any way to improve that?

My first import is normalize.css and after taking a look in the `<style>` element I saw some empty `:root` elements (due to the fact I use future proof CSS transformed via cssnext).
It seems the parser you are using is kind of outdated for some part.

If you are using csso for minification, would you consider switch to a more fresh tool (csswring) based on an parser with an active development (postcss)?

jhnns:
PostCSS is backed by Autoprefixer, so it's probably developed more actively. :+1:

MoOx:
postcss 4 is really fast (just run this to check it out https://github.com/postcss/postcss/tree/master/benchmark ) & even support CSS specs that are not implemented in browsers. So seems a really good choice indeed.

sokra:
Yes, write a PR... I've already tried to use an alternative library in the `rework` branch.
Check that all test pass and maybe add some for the new stuff.

The `rework` branch doesn't work out, because it removed the whitespace from css and this would break SourceMap.

MoOx:
[...]

PR #39, opened 2015-02-03
Add quickfix for csso parse error

jhnns:
This pull request adds a quickfix for the csso parse error.

#36
https://github.com/css/csso/issues/185

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,7 +15,7 @@ module.exports = function(content, map) {
 	var forceMinimize = query.minimize;
 	var importLoaders = parseInt(query.importLoaders, 10) || 0;
 	var minimize = typeof forceMinimize !== "undefined" ? !!forceMinimize : (this && this.minimize);
-	var tree = csso.parse(content, "stylesheet");
+	var tree = csso.parse(prepareContentForCsso(content), "stylesheet");
 	if(tree && minimize) {
 		tree = csso.compress(tree, query.disableStructuralMinification);
 		tree = csso.cleanInfo(tree);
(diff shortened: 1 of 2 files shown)


sokra:
The test doesn't fail without the fix...

It doesn't seem to test the bug... Not sure... Maybe I do something wrong..

jhnns:
Strange... the tests are failing as expected if I just return the content without modifying it

jhnns:
bildschirmfoto 2015-02-03 um 23 30 35
bildschirmfoto 2015-02-03 um 23 30 59
bildschirmfoto 2015-02-03 um 23 31 46

sokra:
> Strange... the tests are failing as expected if I just return the content without modifying it

Only because the atrifical `a{}` is missing...

Construct a test case that demonstrate the bug...

jhnns:
See my first picture. When the content is returned without adding the `a{}`, the tests are failling.

benurb:
I have the same problem and the PR fixed it for me. Thanks :+1:

navinpeiris:
:+1:

sokra:
I would like to reprodude it before merging... So please give me a testcase.

benurb:
I've found the underlying problem in my input files. They were generated by the sass preprocessor, which adds a byte-order-mark to the beginning of the resulting CSS file. csso was unable to parse this file correctly so I opened a PR there: css/csso/pull/223
@sokra: Maybe the BOM-removal-logic is something you want to add to your css-loader, too. But this quickfix here in this PR is not needed imho.

svnlto:
:+1:

sokra:
> @sokra: Maybe the BOM-removal-logic is something you want to add to your css-loader, too. But this quickfix here in this PR is not needed imho.

Maybe we can add this to webpack...

sokra:
> @sokra: Maybe the BOM-removal-logic is something you want to add to your css-loader, too. But this quickfix here in this PR is not needed imho.
> Maybe we can add this to webpack...

We already have this: https://github.com/webpack/core/blob/master/lib/NormalModuleMixin.js#L15-L22

sokra:
Closing as we are no longer using csso...

Issue #29, opened 2014-12-17
sourceMap css does not function for background-image (maybe stylus-loader issue)


nmccready:
So this issue is bizarre and I hope I can make an example that does this.

I successful got sourceMaps to work with stylus-loader with
`{ test: /\.styl$/, loader: 'style!css?sourceMap!stylus' } `

However when `?sourceMap` is enabled all our `background url('whatever/sprite.png')` no longer work. When the ?sourceMap is removed all is well.

I will make an example repo when I get the chance. Also let me know if this should be an issue on stylus-loader instead.

As for more info the css produced are identical except that sourcemap css is redirected via @import some url. The computed CSS reported by Chrome Debugger is identical, but using (Chrome Extension) CSSViewer reports different processed CSS. Which shows the missing sprite.

Also this problem exists in Firefox as well so I do not think it is a chrome issue.

nmccready:
Thank you.

sokra:
> I'll try to find a better solution...

`URL.createObjectURL` + `Blob` + `sourceMappingURL` does work.

nmccready:
Its been a while; how would this look in the config?

sokra:
`style-loader!css-loader?sourceMap`

timaschew:
we are using stylus-loader and all the `url` paths are not resolved, if the stylus file was called via another module via `~myModule`.

With this settings for the stylus-loader it's working: `?{"resolve url": true}`

kevinrenskers:
I just came across the same issue where background images from css don't work because of the paths, when using sourcemaps, css-loader and file-loader in dev mode with webpack-dev-server.

For now I've simply disabled sourcemaps, but was there a better fix? `style-loader!css-loader?sourceMap` was mentioned but this is exactly what's breaking for me.

sokra:
I think a workaround is to use an absolute URL as `output.publicPath`.

nmccready:
#113 related; tried the pubicPath still didn't work.

nmccready:
This issue is also failing with fonts as well. They are being declared as they cannot be decoded in chrome or firefox. Oddly it works fine in safari.

nmccready:
https://github.com/webpack/style-loader/issues/55

tkiethanom:
Is there still no fix for this? Specifying an absolute path for output.publicPath does not seem to work for background-image

sokra:
[...]

PR #40, opened 2015-02-13
Use devtool and sourceMap flag from webpack config

mtscout6:
Instead of needing to apply the sourceMap query parameter generate if a
devtool option is specified in the webpack config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,14 +58,6 @@ The result is:
 
 * `url(/image.png)` => `require("./image.png")`
 
-### SourceMaps
-
-To include SourceMaps set the `sourceMap` query param.
-
-`require("css-loader?sourceMap!./file.css")`
-
-I. e. the extract-text-webpack-plugin can handle them.
-
 ### importing and chained loaders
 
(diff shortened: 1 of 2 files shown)


jhnns:
Seems reasonable... @sokra why did you use a query param?

sokra:
This `sourceMap` is different from SourceMaps in other loaders. It has a runtime overhead. (And is currently broken #29, but this is an other problem).

It injects the SourceMap to the resulting bundle (instead of generating a separate SourceMap file). You can use it without the SourceMap devtool. It is not related to the devtool at all. So I would better leave it opt-in. Currently it's only working correctly with `ExtractTextPlugin` with `allChunks: true`. In other cases it breaks when images/fonts are referenced (because of Cross-Origin-Permission error ... :cry:). We need to find a whole different approach to implement it (or request browser support for `@SourceMappingURL` in `<style>`).

mtscout6:
Closing in favor of webpack/extract-text-webpack-plugin#53

PR #45, opened 2015-04-14
added query parameter url=false

catdad:
This pull request is to address the following feature request.
https://github.com/webpack/css-loader/issues/44


catdad:
As soon as I submitted this, I thought that `ignoreUrl` would be a better query parameter (and does not have to deal with `false` as a string). Easy change though, I can make that if it is more appropriate.

sokra:
see #183

PR #67, opened 2015-06-02
Use lodash.clone to copy query-params for CleanCSS

SimenB:
As discussed on Gitter https://gitter.im/webpack/webpack?at=556d9496463d0c7c066db585

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -8,6 +8,7 @@ var ReplaceMany = require("./ReplaceMany");
 var loaderUtils = require("loader-utils");
 var SourceListMap = require("source-list-map").SourceListMap;
 var CleanCSS = require("clean-css");
+var clone = require("lodash.clone");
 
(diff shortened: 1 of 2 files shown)


SimenB:
@sokra Any thoughts on this? I'd prefer not having to float a patch until a better solution is found (unless it's tomorrow). :smile:

sokra:
closing because using postcss now.

SimenB:
@sokra Could you update the readme? It still points to CleanCSS
https://github.com/webpack/css-loader/blob/master/README.md#minification

SimenB:
@sokra How do I pass options to `cssnano`? I'd like to pass `false` to `autoprefixer`
http://cssnano.co/options/

sokra:
yes, you can do that now with 0.16.0, `css-loader?-autoprefixer`

SimenB:
Awesome, thanks!

PR #68, opened 2015-06-02
support for non url requests in module mode

martnu:


martnu:
build error doesn't seem related to my changes, safe to ignore?

sokra:
The build error means that this change decreases the test coverage... (normal because tests are missing, but that's fine here)

sokra:
https://github.com/webpack/css-loader/commit/11b438ec87213a9f21bdeb4817b8486e4c59df88 this fixes the coverage error.

PR #73, opened 2015-06-09
Improved class and ID regex selector match.

andy-shea:
This fixes the incorrect matching of floating point keyframe values as class names:

    @keyframes z {
      2.5% {color: green;}
    }

Previously, the "5" in "2.5%" would be matched as a class name.  This commit fixes that.


sokra:
Thanks

PR #91, opened 2015-06-29
Disable unused rule removal & identifier reduction. Fixes #89.

### LINKED ISSUE #89: css-loader 0.15.1 breaks keyframe animations
For me it either:

1. Removes "unused" keyframe animations (when keyframe animation is defined in one file but is used in another file only - it just guesses that this keyframes are not used anywhere)
2. If two animations are defined in one file, this file is then included in another two files, animations get the same single letter name `a` which breaks one of them with lower priority (because the latter overrides the former) - see [more details here](http://stackoverflow.com/questions/31070033/broken-css-keyframe-animations-when-using-webpacks-css-loader-with-uglifyjs-plu).

ben-eb:


markdalgleish:
Thanks! :smile:

ben-eb:
No problem. Ping me if you have any other issues with compression (most of the advanced options are designed for a single CSS file). :smile:

sokra:
a testcase would be great

PR #96, opened 2015-07-08
Fix broken composition with class names that contain characters other than [A-Za-z0-9_]

rtsao:
Class composition is broken with "non-word" (`\w`) characters due to the regex used to split tokens.

Perfectly valid classes such as `.my-class` or `.mycläss` are broken as they get split up.

This PR updates the regex to use `\S` instead of `\w`which allows these characters and thus fixes the problem. 

Additionally, a test has been added to verify composition works with a class that contains a hyphen (which fails without the updated regex).

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -20,7 +20,7 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 		var urlItems = [];
 
 		function replaceImportsInString(str) {
-			var tokens = str.split(/(\w+)/);
+			var tokens = str.split(/(\S+)/);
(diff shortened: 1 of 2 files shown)


noamlerner:
This is awesome!

marthakelly:
This would really help out a project we're working on at Uber! Please let us know if we can do anything to help get this merged, or anything you'd like changed before we do that.

+1

freeqaz:
+1

freeqaz:
.über

markdalgleish:
Merged, published as v0.5.14. Thanks for the PR 😀

PR #115, opened 2015-07-22
Adds [dir] and [sourceHash:<len>] in getLocalIdent

Aluxian:
I added [dir] because [local] is always the same for me ("styles") and [path] is too verbose.

[sourceHash] fixes https://github.com/webpack/css-loader/issues/93


Aluxian:
Any thoughts about merging this? What do you think about it?

jedrichards:
I think there's a `[folder]` option. Does that work for you?

Aluxian:
I'm going to close this because it's so outdated. I'm not working on that project anymore, anyway.

PR #116, opened 2015-07-23
Update cssnano to 2.1.0

oliverwoodings:
There is a bug in the version of `postcss-normalize-urls` used by cssnano 1.4.2 which causes it to mess up data URIs (e.g. inline fonts and images) by converting `//` to `/`. Updating to the latest cssnano resolves this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "css loader module for webpack",
   "dependencies": {
     "css-selector-tokenizer": "^0.5.1",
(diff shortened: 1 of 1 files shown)


ben-eb:
:+1:, didn't realise it was using `1.x`.

oliverwoodings:
@ben-eb was a fun bug to track down! Sat there disabling cssnano plugins one by one until it worked again

sokra:
Thanks

PR #130, opened 2015-08-25
Allow ISO 10646 characters U+00A0 and higher in local ident

dtinth:
I am setting `localIdentName=[name]の[local]` and was surprised that the `の` has been changed into a hyphen.

In CSS, any characters above U+00A0 are allowed, [per the specification](http://www.w3.org/TR/CSS21/syndata.html#characters) (emphasis mine):

> In CSS, _identifiers_ (including element names, classes, and IDs in [selectors](http://www.w3.org/TR/CSS21/selector.html)) can contain only the characters [a-zA-Z0-9] __and ISO 10646 characters U+00A0 and higher__, plus the hyphen (-) and the underscore (_) […]


sokra:
Thanks

PR #154, opened 2015-09-30
Fix confusing paths in source maps

jhnns:
Since CSS source maps are presented differently in the browser, we don't need the loader stuff in CSS source map sources. What do you think?

Before fix, paths (with the sass-loader) looked like:
Before

After fix, paths look like:
After

Also fixes jtangelder/sass-loader#150

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -97,10 +97,14 @@ module.exports = function(content, map) {
 			map = result.map;
 			if(map.sources) {
 				map.sources = map.sources.map(function(source) {
-					var p = path.relative(query.context || this.options.context, source).replace(/\\/g, "/");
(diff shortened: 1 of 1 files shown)


jhnns:
@sokra

jhnns:
Still need to test it, but thx anyway :+1:

danauclair:
I'm still seeing a similar issue when using both the autoprefixer-loader and sass-loader, with a webpack config setup like this project: https://github.com/erikras/react-redux-universal-hot-example

The confusing paths are noted in this issue:
https://github.com/erikras/react-redux-universal-hot-example/issues/572

Removing autoprefixer-loader from the chain results in normal looking sourcemap paths. Any ideas?

PR #160, opened 2015-10-07
fixes #150

### LINKED ISSUE #150: processCss ignores props passed to cssnano via loader query
`css-loader?{"sourceMap":true,"minimize":true,"safe":true}`

[lib/processCss.js:157](https://github.com/webpack/css-loader/blob/master/lib/processCss.js#L157) should use `Object.assign` instead of `Object.create`.

```js
var minimizeOptions = Object.create(query);
```
`minimizeOptions` will be:
```json
{}
```

```js
var minimizeOptions = Object.assign({}, query);
```
`minimizeOptions` will be:
```json
{ "sourceMap": true, "minimize": true, "safe": true } 
```

cc @sokra

mdreizin:
Need to add `Object.assign` [polyfill](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign).

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -154,7 +154,7 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
 	]);
 
 	if(minimize) {
-		var minimizeOptions = Object.create(query);
+		var minimizeOptions = Object.assign({}, query);
 		["zindex", "normalizeUrl", "discardUnused", "mergeIdents", "reduceIdents"].forEach(function(name) {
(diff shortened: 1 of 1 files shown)


sokra:
The CI build is broken. You may use the `lodash.assign` npm module instead of Object.assign to make it compatible.

sokra:
I fixed it.

mdreizin:
@sokra Thanks a lot :)

PR #169, opened 2015-10-22
Added missing camelCase option for locals loader

nkt:
Fixes https://github.com/webpack/css-loader/pull/149#issuecomment-150114155


sokra:
Thanks, good work.

PR #183, opened 2015-11-13
Resolved  https://github.com/webpack/css-loader/issues/102

ShyykoSerhiy:


lib/loader.js
```diff
@@ -64,23 +64,28 @@ module.exports = function(content, map) {
...
-			if(idx > 0) { // idx === 0 is catched by isUrlRequest
-				// in cases like url('webfont.eot?#iefix')
-				urlRequest = url.substr(0, idx);
-				return "\" + require(" + loaderUtils.stringifyRequest(this, urlRequest) + ") + \"" +
-					url.substr(idx);
-			}
-			urlRequest = url;
-			return "\" + require(" + loaderUtils.stringifyRequest(this, urlRequest) + ") + \"";
-		}.bind(this));
+		if(!query.hasOwnProperty('import')) {
```
sokra: Please don't check for exisitance of query properties. Check for the value. `query.import !== false`. Document it as `css-loader?-url&-import`.

Elsewise this cases would be broken:

* `css-loader?url`
* `css-loader?{url: true}`

lib/processCss.js
```diff
@@ -83,23 +86,27 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
...
+								item.stringType = "";
+								delete item.innerSpacingBefore;
+								delete item.innerSpacingAfter;
+								var url = item.url;
+								item.url = "___CSS_LOADER_URL___" + urlItems.length + "___";
+								urlItems.push({
+									url: url
+								});
+							}
+							break;
```
sokra: The break should be outside of the `if`, at the end of the case clause.

sokra:
Please add test cases that validate the behavior.

ShyykoSerhiy:
@sokra I've added some tests and fixed checks

jasonslyvia:
Please merge this!

xile611:
Please merge this!

arcthur:
Please merge this!

jasonHzq:
Please merge this!

shelbeniskb:
please merge this!

sokra:
> please merge this!

Instead of spamming `please merge this!` try to review the PR and comment with broken stuff. I. e. here were test regarding the `@import` handling missing. And no tests propably means: behavior is broken.

vikalpgupta-birdeye:
@sokra This parameter works for url rewriting inside the stylesheet, but it also does not emit the files, can there be a parameter, in which we can specify such requirement

PR #187, opened 2015-11-14
rewrite sourceMap file suffix to css for sass or less

jetiny:
sourceMap name should not be same as .sass file


jetiny:
image

michael-ciniawsky:
@jetiny What's your usecase? :) Please rebase, close and reopen the PR to trigger the CLA Bot again and recap on actual use 😛 thx in advance

michael-ciniawsky:
Thx, but this shouldn't be done within a loader, if you the original file has a `extname !== .css` rhis is the correct behaviour

PR #202, opened 2015-12-08
Support for nested items

### LINKED ISSUE #200: CSS loader not extracting path in -webkit-image-set
Hi,

we currently use this mixin in our LESS code to generate retina display background images 

```
.retina-image(@width, @height, @normal-image, @retina-image) {
  width: @width;
  height: @height;
  background-image: url(@normal-image); // fallback for Firefox and IE
  background-image: -webkit-image-set(
    url(@normal-image) 1x,
    url(@retina-image) 2x
  );
  background-repeat: no-repeat;
  display: inline-block;
}
```
and it seems like the url in `background-image: url(@normal-image);` is picked up and replaced by the loader but not the ones in `-webkit-image-set`. They are just rendered as `-webkit-image-set(url('/path/to/image') 1x, url('/path/to/image@2x'))`.

lemieux:
This allows value like `-webkit-image-set(url('some-url.png') 1x, url('some-other-url.png') 2x)` to be parsed and transformed the by loader.

It fixes #200.

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -78,29 +78,29 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 		});
 
 		function processNode(item, isValue) {
-            switch (item.type) {
+			switch (item.type) {
 				case "value":
-                    item.nodes.forEach(function(node){
+					item.nodes.forEach(function(node){
 						processNode(node, true);
 					});
-                    break;
-                case "nested-item":
-                    item.nodes.forEach(function(node){
+					break;
+				case "nested-item":
+					item.nodes.forEach(function(node){
 						processNode(node);
 					});
-                    break;
-                case "item":
-                    var importIndex = imports["$" + item.name];
-                    if (typeof importIndex === "number") {
-                        item.name = "___CSS_LOADER_IMPORT___" + importIndex + "___";
-                    }
-                    break;
-                case "url":
-                    if (!/^#/.test(item.url) && loaderUtils.isUrlRequest(item.url, options.root)) {
-                        item.stringType = "";
-                        delete item.innerSpacingBefore;
-                        delete item.innerSpacingAfter;
-                        var url;
+					break;
+				case "item":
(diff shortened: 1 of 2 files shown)


lib/processCss.js
```diff
@@ -77,30 +77,55 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
...
+                    }
+                    break;
+                case "url":
+                    if (!/^#/.test(item.url) && loaderUtils.isUrlRequest(item.url, options.root)) {
+                        item.stringType = "";
+                        delete item.innerSpacingBefore;
+                        delete item.innerSpacingAfter;
+                        var url;
+
+						if(isValue) {
```
sokra: For what reason is `isValue` tracked?

sokra:
please fix the spacing. It should be tabs.

lib/processCss.js
lemieux: maybe it should be named differently, but it's basically to reuse the same `processNode` logic on nested properties... should probably be named `isNested`

lib/processCss.js
```diff
@@ -77,30 +77,55 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
...
+					}
+					break;
+				case "url":
+					if (!/^#/.test(item.url) && loaderUtils.isUrlRequest(item.url, options.root)) {
+						item.stringType = "";
+						delete item.innerSpacingBefore;
+						delete item.innerSpacingAfter;
+						var url;
+
+						if(isValue) {
```
sokra: `isValue` is always `true` here, because a nested item has this representation:

``` js
{
	type: "nested-item",
	name: name,
	nodes: [
		{ type: "value", nodes: [...] }
	]
}
```

So I think you could remove this check here and always use the true case.

lemieux: using the true case makes the background img 2 test and the font face tests fail.

lemieux: If I remember correctly, it was because in the case where it's processing a value, the urls in it were not transformed.

Issue #144, opened 2015-09-23
Promise is not defined error in 0.19.0


chrismcband:
Latest release is failing builds for me, seeing the error below in 0.19.0. This error is not present in 0.18.0.

```
ERROR in ./~/css-loader!./app/main.css
Module build failed: ReferenceError: Promise is not defined
    at LazyResult.async (/Users/chris/IdeaProjects/SW2.0/survivejs/node_modules/css-loader/node_modules/postcss/lib/lazy-result.js:152:31)
    at LazyResult.then (/Users/chris/IdeaProjects/SW2.0/survivejs/node_modules/css-loader/node_modules/postcss/lib/lazy-result.js:75:21)
    at processCss (/Users/chris/IdeaProjects/SW2.0/survivejs/node_modules/css-loader/lib/processCss.js:174:5)
    at Object.module.exports (/Users/chris/IdeaProjects/SW2.0/survivejs/node_modules/css-loader/lib/loader.js:22:2)
 @ ./app/main.css 4:14-75 13:2-17:4 14:20-81
```

Here's the webpack config:
```javascript
var path = require('path');
var HtmlwebpackPlugin = require('html-webpack-plugin');
var webpack = require('webpack');

var ROOT_PATH = path.resolve(__dirname);

module.exports = {
  entry: path.resolve(ROOT_PATH, 'app'),
  output: {
    path: path.resolve(ROOT_PATH, 'build'),
    filename: 'bundle.js'
  },
  devtool: 'eval-source-map',
  devServer: {
    historyApiFallback: true,
    hot: true,
    inline: true,
    progress: true
  },
  module: {
    loaders: [
      {
        test: /\.css$/,
        loaders: ['style', 'css'],
        include: path.resolve(ROOT_PATH, 'app')
      }
    ]
  },
  plugins: [
    new webpack.HotModuleReplacementPlugin(),
    new HtmlwebpackPlugin({
      title: 'Kanban app'
    })
  ]
};
```

And package.json:
```json
{
  "name": "survivejs",
  "version": "0.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "start": "webpack-dev-server",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "css-loader": "^0.19.0",
    "html-webpack-plugin": "^1.6.1",
    "node-libs-browser": "^0.5.3",
    "style-loader": "^0.12.4",
[...]

PR #218, opened 2016-01-11
Specify node engine minimum version

dreyescat:
Setting the minimum version of the node engine will make npm to warn or
complain, if node `engine-strict` config is set to true, whenever the
installation environment does not meet this requirement.

Background:

`css-loader` depends on `postcss` that uses native promises. Promises
where introduced in node version 0.12.0. So `css-loader`
*transitively inherits* this dependency.

Related to #144.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "0.23.1",
   "author": "Tobias Koppers @sokra",
   "description": "css loader module for webpack",
+  "engines": {
+    "node": ">=0.12.0"
+  },
   "dependencies": {
     "css-selector-tokenizer": "^0.5.1",
     "cssnano": ">=2.6.1 <4",


dreyescat:
It looks like that forcing [engineStrict](https://docs.npmjs.com/files/package.json#enginestrict) per package was deprecated since `npm` version `3.0.0`.
The user needs to explicitly specify in config if `npm will stubbornly refuse to install` setting the [`engine-strict`](https://docs.npmjs.com/misc/config#engine-strict) option to true. Otherwise a simple warning will be shown.

With this change using a node version below `0.12.0`
* with `engine-strict` mode set to `true`  we will get:
```
npm ERR! notsup Unsupported
npm ERR! notsup Not compatible with your version of node/npm: css-loader@0.23.1
npm ERR! notsup Required: {"node":">=0.12.0"}
npm ERR! notsup Actual:   {"npm":"1.4.28","node":"0.10.40"}
```

* with `engine-strict` set to  `false` we will get:
```
npm WARN engine css-loader@0.23.1: wanted: {"node":">=0.12.0"} (current: {"node":"0.10.40","npm":"1.4.28"})
npm WARN engine postcss@5.0.14: wanted: {"node":">=0.12"} (current: {"node":"0.10.40","npm":"1.4.28"})
npm WARN engine csso@1.4.4: wanted: {"node":">=0.12.0"} (current: {"node":"0.10.40","npm":"1.4.28"})
```

As you can see `postcss` and `csso` dependencies already have this requirement and probably should be enough because they are actually the ones that introduced this requirement (use node native Promises). So it might be fair just to let them warn of this requirement and not `css-loader` who doesn't seem to need this requirement per se.

But... If you set `engine-strict` to `true` to the version without this change, then you get a really confusing message:

```
npm ERR! notarget No compatible version found: postcss@'>=5.0.10-0 <6.0.0-0'
npm ERR! notarget Valid install targets:
npm ERR! notarget ["0.1.0","0.2.0","0.3.0","0.3.1","0.3.2","0.3.3","0.3.4","0.3.5","1.0.0","2.0.0","2.1.0","2.1.1","2.1.2","2.2.0","2.2.1","2.2.2","2.2.3","2.2.4","2.2.5","2.2.6","3.0.0","3.0.1","3.0.2","3.0.3","3.0.4","3.0.5","3.0.6","3.0.7","4.0.0","4.0.1","4.0.2","4.0.3","4.0.4","4.0.5","4.0.6","4.1.0","4.1.1","4.1.2","4.1.3","4.1.4","4.1.5","4.1.6","4.1.7","4.1.8","4.1.9","4.1.10","4.1.11","4.1.12","4.1.13","4.1.14","4.1.15","4.1.16","5.0.0","5.0.1","5.0.2","5.0.3","5.0.4","5.0.5","5.0.6"]
```

So, even this is not a `css-loader` issue maybe it is good to keep just to get better feedback to end users (compare the two outputs when `engine-strict` is set to `true`).

Feel free to reject this PR if you think it is not going to be helpful :wink:.

sokra:
Thanks

PR #242, opened 2016-03-10
add documentation for camelCase query parameter

ianmcnally:
Hi @sokra et al,

I was searching all around the internet trying to find a way to camelize the class names exported from css loader (in conjunction with css modules), when I found your commit adding the camelCase query. :clap: 

I was hoping to surface that information in the documentation.


sokra:
Thasnk

PR #274, opened 2016-05-06
Add alias feature to rewrite urls

bbtfr:
Hi guys,
I found sometimes, it's not easy to change the source file / input file of css-loader, e.g.: when we're using sass files in another package (`bootstrap-sass`), we'll get a lot of complains about `Cannot resolve 'file' or 'directory' ...`. 
There are tons of questions on Google about this. 
For `bootstrap`, we may use `$icon-font-path` or `bootstrap-loader`, but for other projects, we have nothing.
`resolve-url-loader` is a solution, but it's based on source-map and makes the compilation much slower.
So what do you think if we rewrite urls in `css-loader`?

Here's an example to load `bootstrap-sass` with `sass-loader`, `css-loader` & `style-loader`

``` javascript
// webpack.config.js
// ## Webpack Configuration
var path = require('path');

module.exports = {
  entry: "./src/entry",
  output: {
    filename: 'bundle.js',
    path: path.resolve('./dist')
  },
  module: {
    // ### Loaders
    loaders: [
      { test: /\.js$/, loader: "babel" },
      { test: /\.scss$/, loader: "style!css!sass" },
      { test: /\.(woff|woff2|ttf|eot|svg)$/, loader: "url?limit=10000" }
    ]
  },
  cssLoader: {
    alias: {
      "../fonts/bootstrap": "bootstrap-sass/assets/fonts/bootstrap"
    }
  },
  sassLoader: {
    includePaths: [
      path.resolve("./node_modules/bootstrap-sass/assets/stylesheets")
    ]
  }
};
```

``` sass
// stylesheet.scss
@charset "UTF-8";
@import "bootstrap";
```

`cssLoader.alias` use the same syntax with `resolve.alias`, https://webpack.github.io/docs/configuration.html#resolve-alias

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ To be compatible with existing css files (if not in CSS Module mode):
 |**`sourceMap`**|`false`|Enable/Disable Sourcemaps|
 |**`camelCase`**|`false`|Export Classnames in CamelCase|
 |**`importLoaders`**|`0`|Number of loaders applied before CSS loader|
+|**`alias`**|`{}`|Create aliases to import certain modules more easily|
 
 The following webpack config can load CSS files, embed small PNG/JPG/GIF/SVG images as well as fonts as [Data URLs](https://tools.ietf.org/html/rfc2397) and copy larger files to the output directory.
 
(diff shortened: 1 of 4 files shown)


d3viant0ne:
This will need to be updated to 2.x syntax before it's reviewed. With the defaults updates requiring major versions, this would have to be compatible / focused on Webpack > 2.0

bebraw:
Can you sign the CLA and document the feature at readme?

bbtfr:
The code has been rebased, updated to 2.x syntax
The alias resolving functionality has been moved to `createResolver.js`, so it would be clearer
And the document is also been updated

PR #284, opened 2016-06-07
Avoid name collision

themouette:
When 2 files from different projects share the same relative file path AND
selector, the generated hash is the same.

This should not happen, as css-modules are designed to avoid collision.

This pull request add the file content to the hash generation seed, it seems to be the simplest way to do it.

### DIFF
--- a/lib/getLocalIdent.js
+++ b/lib/getLocalIdent.js
@@ -9,7 +9,7 @@ module.exports = function getLocalIdent(loaderContext, localIdentName, localName
 	if(!options.context)
 		options.context = loaderContext.options && typeof loaderContext.options.context === "string" ? loaderContext.options.context : loaderContext.context;
 	var request = path.relative(options.context, loaderContext.resourcePath);
-	options.content = options.hashPrefix + request + "+" + localName;
+	options.content = options.hashPrefix + request + "+" + localName + "+" + options.content;
 	localIdentName = localIdentName.replace(/\[local\]/gi, localName);
 	var hash = loaderUtils.interpolateName(loaderContext, localIdentName, options);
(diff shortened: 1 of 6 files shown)


themouette:
I rewrote the commits as I used spaces instead of tabs.

themouette:
@sokra et al, without being rude, any chance you get an eye on this?

sokra:
I think there was a reason why the hash doesn't include the content hash. I should be calculatable with only path and name. I. e. css-modules-require-hook relies on the current behavior.

But there is a `hashPrefix` option, which allows to modify the hash per project.

joshwnj:
One reason to not include the content hash is because it greatly broadens the scope of things that need to be regenerated, when using watch-mode or hot reloading. For example, if you make a change to a css file, that changes the content hash and therefore all of the classnames exported by that module.

So not impossible, it just requires the computer to do a lot more work which might not be necessary, especially given it seems the requirements can be met with `hashPrefix` as @sokra suggested

themouette:
How could I have missed the `hashPrefix` option?

Thanks, that made my day!

navgarcha:
Would this PR not the issue with the following scenario:

- You have 2 entry chunks.
- Both compose some common classes.
- One overwrites some rules within the class it composes.
- You add both of these to a webpack page (let's say they're some sort of widget).

Here's where the problem occurs. Depending on the order in which these two widgets are added to the page, the original composed rule will overwrite the 'extended' changes. If there was a way to add the entry chunk name to the localIdentName we could avoid this as the classnames would be unique to the entry chunk and therefore cause no conflict - currently there is no way to achieve this or am i missing something (I dont think `hashPrefix` resolves this...tbh i can't seem to get that to do anything at all)?

PR #287, opened 2016-06-09
refactor: manage loader resolution with webpack

### LINKED ISSUE #286: CSS Module composes imports aren't added to webpack
Say there's a project like this:

``` css
/* a.css */
.red { color: red };

/* b.css */
.main { composes: red from './a.css'; }
```

and a config roughly like:

``` js
module: {
  loaders: [
    {
      test: /a\.css$/,
      loader: 'style-loader!css-loader?modules&localIdentName=a--[local]',
    },
    {
      test: /b\.css$/,
      loader: 'style-loader!css-loader?modules&localIdentName=b--[local]',
    },
  ]
}
```

`a.css` will be processed by `b.css`' loader config even though it does't match the file test. This results in the equivalent of:

``` css
.b--red { color: red; }
.b--main { };
```

even though I'd expect the output to look like:

``` css
.a--red { color: red; }
.b--main { };
```

Ideally `css-loader` should `webpack.addDependency` for css module source dependencies so they get processed with the same set of loaders as the rest of the config and not jus the root file's loader url.

chrislloyd:
Enables the behavior detailed in #286.

Instead of assuming that child-loaders are the same as the parent, use Webpack’s build in dependency resolution. The benefits of this is that it’s easier to treat the CSS dependency tree differently (i.e. extract
text for certain vendor libs). The tradeoff is that it gives people the ability to shoot themselves in the foot if they misconfigure their loaders. This is worthwhile IMHO as the error will fail fast.

It shouldn’t change behavior for existing users.


geelen:
So if you've only got one loader matching all your CSS files, this will have no impact, right?

chrislloyd:
That's correct.

delijah:
Could we go forward with this pull request?

sokra:
Currently this doesn't work for the common case where you match `style-loader` and `css-loader` to `*.css`.

But will probably have a better experience with webpack 2 here.

For the webpack 2 and the next version of css-loader we can recommend this ruleset:

``` js
rules: [
  { test: /\.css$/, rules: [
    { issuer: /\.js$/, use: "style-loader" },
    { use: "css-loader" }
  ] },
  { test: /\.sass$/, rules: [
    { issuer: /\.js$/, use: "style-loader" },
    { use: ["css-loader", "sass-loader"] }
  ] }
]
```

With this rules you can even `@import` sass from css...

michael-ciniawsky:
Fixes #286

michael-ciniawsky:
Closed due to inactivity, feel free to reopen :)

PR #293, opened 2016-06-28
fix(processCss/processNode): don't support AlphaImageLoader(src='/xxx')

netwjx:


sokra:
Please add test cases for this behavior.

michael-ciniawsky:
@netwjx What exactly breaks 'AlphaImageLoader', can you elaborate on this case in a more generic way ? 😛 If this should be a fix for a single uncommon loader alone, I have to reject this PR

michael-ciniawsky:
@netwjx I'm interpreting the 👍  as 'yep, this PR is a specific change' and close the PR for now, if my assumption should be wrong, feel free to reopen 😛

Issue #133, opened 2015-09-03
Escaped characters in CSS content are not respected


chandlervdw:
Even though this was listed in #87 and supposedly fixed by https://github.com/css-modules/css-selector-tokenizer/commit/5aa6180c27e1444d4f9601ec5f7ad6b3e894b587, I'm still having this issue.

`content: "\A";` is compiled to `content: "A";`, so the `\` character isn't being parsed correctly...

chandlervdw:
@andreypopp @JackVCurtis @mathieumg are any of you still able to reproduce this issue?

chandlervdw:
I have no idea why this worked but putting ```content: '\"\A';``` fixed it. It outputs as ```content: '\"\A';``` but somehow still renders as a line break. I'm sure this will break soon...

chandlervdw:
Ok, nevermind, that does NOT work.

robmcguinness:
`content: '\0000a0';` is being converted to `content: 'A 0';`.  Downgrading to `v0.14.5` fixes the issue for me.

chandlervdw:
Any word on this? I hate keeping this lib at `v0.14.5`.

robmcguinness:
@chandlervdw my workaround was to alter content value to work with `v0.14.5`+

![image](https://cloud.githubusercontent.com/assets/281070/12646543/bac9cd94-c59e-11e5-84bc-f8a1db9e889e.png)

chandlervdw:
@robmcguinness I'm trying to achieve an end value of `\A` ... can you advise me on what I should change my content value to?

robmcguinness:
@chandlervdw I haven't dug into the code in a while but if I get a minute i'll take a peek

pieter:
Traced the problem to https://github.com/css-modules/css-selector-tokenizer/issues/12

h6ah4i:
+1

kencana-voilaah:
hi guys, encountering same issue, any tweak i should do to make it work for now?

tadjohnston:
Added a PR to bump the`css-selector-tokenizer` version to `0.6.0`.

PR #307, opened 2016-07-25
Fixes issue #133 with unicode strings

tadjohnston:
https://github.com/webpack/css-loader/issues/133

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "node": ">=0.12.0"
   },
   "dependencies": {
-    "css-selector-tokenizer": "^0.5.1",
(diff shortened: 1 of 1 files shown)


TheLarkInn:
@sokra this looks good but I wonder if maybe these tests should be updated to reflect the same node versions as webpack/webpack. We can get back around to it if we need to, I'll assign to me to keep track.

sokra:
Thanks

@TheLarkInn yep

MicheleBertoli:
@sokra it would be awesome to get this on `npm`, thanks! :)

ricsv:
+1 on `npm`, would be really great!

PR #311, opened 2016-07-29
refactor: remove CSS modules in global mode

### LINKED ISSUE #308: [Discussion] Deprecate :local
In my mind, it makes more sense to consider local-scope (and CSS Modules by extension) as something that's **always on or always off**.

I'm seeing a lot of folks hitting the problem where they enable `?modules` and then want to include non-local CSS (e.g. https://github.com/webpack/css-loader/issues/253) but never in the same file. I think @chrislloyd's suggestion in https://github.com/webpack/css-loader/pull/287 is a better approach, so you can configure two instances of `css-loader`, one local and one global, maybe having ".module.css" be local and ".css" be global (like https://github.com/gatsbyjs/gatsby/issues/193). 

In global CSS files, then, I don't see any reason to keep `:local`, except that that was how CSS Modules got started in the first place. But I'm interested in having a discussion about this because I want to know if anyone's still using `:local` and feels it should be kept.

geelen:
Fixes #308

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -147,33 +147,40 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
 		import: query.import !== false
 	};
 
-	var pipeline = postcss([
-		localByDefault({
-			mode: options.mode,
-			rewriteUrl: function(global, url) {
-				if(parserOptions.url){
-					if(!loaderUtils.isUrlRequest(url, root)) {
-						return url;
+	var pipeline = postcss(
+		(options.mode === 'local' ? [
+			/* LOCAL ONLY POSTCSS PLUGINS */
+			localByDefault({
+				mode: options.mode,
+				rewriteUrl: function(global, url) {
+					if(parserOptions.url){
+						if(!loaderUtils.isUrlRequest(url, root)) {
+							return url;
+						}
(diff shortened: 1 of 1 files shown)


lib/processCss.js
```diff
@@ -147,34 +147,41 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
...
-			mode: options.mode,
-			rewriteUrl: function(global, url) {
-				if(parserOptions.url){
-					if(!loaderUtils.isUrlRequest(url, root)) {
-						return url;
-					}
-					if(global) {
-						return loaderUtils.urlToRequest(url, root);
+	var pipeline = postcss(
+		(options.mode === 'local' ? [
```
sokra: I think there is also a `"pure"` mode which need the postcss plugins too.

SpaceK33z:
@geelen, are you still interested in fixing this?

michael-ciniawsky:
@geelen @sokra What's the status here? 😛

michael-ciniawsky:
@geelen friendly ping, this one is actually really important && appreciated

michael-ciniawsky:
Closed due to inactivity, eventually superseded by #497, feel free to reopen :)

PR #315, opened 2016-08-08
add global options `cssLoader` for query

helloyou2012:
like `babel-loader` :  https://github.com/babel/babel-loader/blob/master/index.js#L36

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -3,6 +3,7 @@
 	Author Tobias Koppers @sokra
 */
 var path = require("path");
+var assign = require('object-assign');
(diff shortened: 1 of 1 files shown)


Chaptykov:
Are there any news about merging this PR?

SpaceK33z:
As of webpack v2, this is no longer relevant. It is now possible to specify an `options` object for each loader. Also see the [upgrade guide](https://webpack.js.org/how-to/upgrade-from-webpack-1/#module-loaders-is-now-module-rules) for some examples.

SpaceK33z:
See also #348.

PR #319, opened 2016-08-15
Consistent format for syntax errors.

andreypopp:
I used babel-code-frame package formatting syntax errors code frames. That makes it consistent with Babel output which is I think very important as it makes it less cognitive efforts to recognize and process the error.

You can also see that there's no duplicated filename in error message as webpack already print module name.

Before:

screenshot 2016-08-15 19 47 37

After:

screenshot 2016-08-15 19 46 07

See also:
* https://github.com/babel/babel-loader/pull/287
* https://github.com/postcss/postcss-loader/pull/87

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -2,6 +2,7 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
+var formatCodeFrame = require("babel-code-frame");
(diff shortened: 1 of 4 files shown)


sokra:
awsome :+1:. A test would be great :smile:

andreypopp:
@sokra done!

sokra:
Thanks

PR #321, opened 2016-08-23
The way I should specify minification options is "unpleasant"

termosa:
I will disfigure my configuration file if I try to specify many options for `cssnano`. Why not provide an option to parse the file with options?

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,24 @@ This may change in the future, when the module system (i. e. webpack) supports l
 
 By default the css-loader minimizes the css if specified by the module system.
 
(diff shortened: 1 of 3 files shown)


andreypopp:
Not that this is also possible to configure using `query` key in loader config:
```
+module.exports = {
 +  ...,
 +  module: {
 +    loaders: [
 +      { test: /\.css$/, loader: "css", query: {minimize: true} }
 +    ]
 +  }
 +};
```

termosa:
@andreypopp wow, now I know a bit more about webpack.
But still, it requires me to load configurations by my own and to use object-assign…
Are you sure you don't want to provide this option? Or at least support of `cssnano` options in webpack config like eslint did?

andreypopp:
@termosa idk, not a maintainer, but my opinion is that it's better to keep the loader implementation simpler. Also I think eslint-loader shouldn't support reading from `.eslintrc`.

sokra:
I agree with @andreypopp. `options` in the configuration should be with only source of loader settings.

PR #347, opened 2016-09-28
Fix #340

### LINKED ISSUE #340: Incompat with webpack2 LoaderOptionsPlugin
In lib/loader.js
[https://github.com/webpack/css-loader/blob/master/lib/loader.js#L101](url)
`this.options` may be rewritten by LoaderOptionsPlugin in the following line
[https://github.com/webpack/webpack/blob/master/lib/LoaderOptionsPlugin.js#L29](url)
thus `this.options.context` could be undefined and cause an error.
Should we change it to `this.context` to fix the bug?

tomgeorge:
this is a fix for #340.

The WebpackLoaderOptions plugin appears to overwrite `this.options`, causing `this.context.options` to be undefined and throwing `path should be a string` errors.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -98,7 +98,7 @@ module.exports = function(content, map) {
 			if(map.sources) {
 				map.sources = map.sources.map(function(source) {
 					source = source.split("!").pop();
-					var p = path.relative(query.context || this.options.context, source).replace(/\\/g, "/");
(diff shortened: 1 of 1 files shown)


tomgeorge:
Closing because it failed tests and does not appear to be an issue on my work machine anymore

Alxandr:
I'm still having this issue.

PR #327, opened 2016-09-06
Pass `customInterpolateName` from query

Chaptykov:
It is based on top of #315 by @helloyou2012.

This PR allows to pass `customInterpolateName` function from webpack config to loaderUtils.interpolateName. It allows to generate classnames manually. It will solve #292. 

Webpack config:

``` js
let i = 0;

module.exports = {
  module: {
    loaders: [{
      test: /\.css$/,
      loader: 'style!css?modules&importLoaders=1&localIdentName=[name]__[local]'
    }]
  },
  cssLoader: {
    customInterpolateName: function() {
      i++;
      return '_' + i;
    }
  }
};
```

For this file structure:

```
first.css
  .title {}
  .text {}
second.css
  .title {}
```

returns

``` css
._1 {} /* first.css  → .title */
._2 {} /* first.css  → .text  */
._3 {} /* second.css → .title */
```

PR #348, opened 2016-09-28
Add support for complex config options

mdreizin:
This PR adds ability to use use custom loader options:

``` js
{
    module: {
        loaders: [{
            test: /\.css$/,
            loaders: [
                'style',
                'css?config=customCssConfig'
            ]
        }]
    },
    customCssConfig: {}
}
```

Instead of #327, #315 I prefer to use recommended way to retrieve the loader config [`loader-utils#getloaderconfig`](https://github.com/webpack/loader-utils#getloaderconfig).

@sokra Could you please review this PR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,36 @@ module.exports = {
 };
 ```
 
+## Advanced options
+
+If you need to pass [more advanced options](https://github.com/webpack/css-loader/pull/348), especially those which cannot be stringified, you can also define an `cssLoader`-property on your `webpack.config.js`:
+
+``` javascript
+module.exports = {
+  module: {
+    loaders: [{
+      test: /\.css$/,
+      loader: "style-loader!css-loader"
+    }]
+  },
+  cssLoader: {}
+};
+```
+
+If you need to define two different loader configs, you can also change the config's property name via `css?config=otherCssLoaderConfig`:
+
+``` javascript
+module.exports = {
+  module: {
+    loaders: [{
+      test: /\.css$/,
+      loader: "style-loader!css-loader?config=otherCssLoaderConfig"
+    }]
+  },
+  otherCssLoaderConfig: {}
+};
+```
+
 ### 'Root-relative' urls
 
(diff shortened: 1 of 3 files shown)


mdreizin:
cc @jhnns @bebraw

bebraw:
What's the benefit over [options](http://webpack.js.org/how-to/upgrade-from-webpack-1/#module-loaders-is-now-module-rules)?

mdreizin:
@bebraw Here is some of benefits:

- it gives you ability to have shared config for group of `xxx-loader`
- it gives you ability override default options without modifying `query`/`options` of  `xxx-loader`
- it gives you ability to have custom configs for some user cases
- it gives you ability to use recommended way to retrieve config of `xxx-loader`
- etc

Please see:

https://github.com/webpack/less-loader#less-plugins
https://github.com/jtangelder/sass-loader#sass-options
https://github.com/webpack/html-loader#advanced-options
https://github.com/webpack/loader-utils/commit/7e86a18035633ffe041bfe43451e3e4352ee13c8

bebraw:
Gotcha. Looks like a good idea then so 👍 from me. I think @sokra can comment on the code once he has some time available.

mdreizin:
@bebraw Thanks a lot.
It would be nice to have this feature as early as possible.
I'm looking forward to get feedback from @sokra.

jhnns:
I thought we agreed that polluting the webpack config object is not a good idea? We also [recently added config object validation to webpack@2](https://github.com/webpack/webpack/pull/2974), so this will not work with webpack 2, right?

bebraw:
@jhnns Ok, I guess in that case we might to remove the feature from those other loaders to keep this consistent and document how to achieve the same effect without.

sokra:
No longer allowed in webpack 2 as @jhnns said

> it gives you ability to have shared config for group of xxx-loader

``` js
const sharedOptions = { ... }
```

> it gives you ability override default options without modifying query/options of xxx-loader

``` js
options: {
  ...sharedOptions,
  key: value
}
```

PR #356, opened 2016-10-10
fix empty context when LoaderOptionsPlugin is used as it wipes the context option

pospi:


pospi:
fixes #340

sokra:
It's only a workaround

PR #358, opened 2016-10-13
Added the possibility of using -url and root options at the same time

alexisgahon:
I was trying to use -url and root together to be able to replace my assets paths without being resolved by webpack.
But these options wasn't working together, so I wasn't able to have my assets resolved by navigators when using sourceMaps (style-loader embed styles inside a blob file).


alexisgahon:
@sokra can you check this ?

sokra:
I don't think this works as expected.

If you think it does work: add tests.

PR #361, opened 2016-10-17
Disable Autoprefixer in cssnano

### LINKED ISSUE #281: Remove Autoprefixer from cssnano
`css-loader` contains cssnano to compress CSS. cssnano by default contains Autoprefixer to remove all prefixes. It works great when you undertand what you want to do. But because css minifier feature is hidden inside webpack many people don’t expect that webpack will remove prefixes.

Right now I have so many issues in Autoprefixer repo about this issue.

So I suggest to remove `autoprefixer` rule from `cssnano` in `css-loader`.

@ben-eb do you agree as cssnano author?

@sokra we have problem right now, because in latest Can I Use update iOS 8 was removed from default `last 2 versions` browsers list and many people are in pain.

ai:
Fix #281

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -179,7 +179,7 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
 
 	if(minimize) {
 		var minimizeOptions = assign({}, query);
-		["zindex", "normalizeUrl", "discardUnused", "mergeIdents", "reduceIdents"].forEach(function(name) {
+		["zindex", "normalizeUrl", "discardUnused", "mergeIdents", "reduceIdents", "autoprefixer"].forEach(function(name) {
 			if(typeof minimizeOptions[name] === "undefined")
 				minimizeOptions[name] = false;
 		});


SpaceK33z:
@jhnns, do you think this change is fine? Apparently this is causing many many issues when using postcss.

jhnns:
Yes, I think it's fine. I would consider this to be a bug, so this is not a major version bump. We (probably) won't break anything by *adding prefixes*, we will just make CSS files bigger (assuming that autoprefixer is working correctly 😉).

jhnns:
And I assume that *most developers* were not aware of this (including me).

ben-eb:
nb: autoprefixer in cssnano is **removal-only**, so it would have not added prefixes, only removed them. 😄

ai:
Thanks. Very excited about this fix. Could you ping me on release, so I could tweet about this important fix for Autoprefixer users?

jhnns:
> nb: autoprefixer in cssnano is removal-only, so it would have not added prefixes, only removed them. 

I know. But when we remove the autoprefixer from cssnano, we will essentially *add prefixes* to minified bundles :)

ben-eb:
Ah yes, of course. 👍

codeheroics:
Thanks for the great work! Is a release containing this fix soon planned for css-loader?

PR #377, opened 2016-11-20
feat: add `asString` option

notVitaliy:
Feature. Inject an actual string export into webpack bundle.

Yes.

Yes.

**Summary**
I had a requirement for this loader to return the value of `cssAsString`. This can prove to be useful for chaining loaders after css-loader that require the css string as the content.

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,6 +275,51 @@ Usage:
 require('file.css').className
 ```
 
+## As String
+
+By default the loader injects a function into the webpack bundle that exports the compiled css as a string. The `asString` query param will export the compiled css as an actual string into the bundle.
+
+Example:
+
+```javascript
+/* 0 */
+/***/ function(module, exports, __webpack_require__) {
+
+exports = module.exports = __webpack_require__(1)();
+// imports
+
+
+// module
+exports.push([module.i, ".clear{ clear: both; }", ""]);
+
+// exports
+
+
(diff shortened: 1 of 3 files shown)


TheLarkInn:
What do you think @jhnns. This could obsolete the need for to-string-loader. Looks like the change is not too heavy. Thoughts?

gdi2290:
devs also use `'exports-loader?module.exports.toString()'`

notVitaliy:
@TheLarkInn to-string-loader behaves differently. It'll output a function that'll return the string in the bundle whereas this feature will output the actual string in the bundle, which is useful for chaining loaders after css-loader that expect an actual string.

michael-ciniawsky:
@notVitaliy  Please sign the CLA by closing and reopening the PR to trigger the CLA Bot again and rebase against current master

michael-ciniawsky:
`asString` => `toString`

But I need to check again, if we don't have any other possibilities to avoid options like that, not because it's a bad idea etc but things like this shouldn't be necessary in general 😛

bebraw:
Yeah, it feels like decomposing css-loader would work better. Better to have a collection of smaller loaders over single scary one.

michael-ciniawsky:
Agreeing with @bebraw this needs to be better solved in `toString-loader`  thx

PR #383, opened 2016-12-05
Bump source-map-list to 0.1.7

gaearon:
Per https://github.com/facebookincubator/create-react-app/pull/1101#issuecomment-264865954, we would like to make sure https://github.com/webpack/source-list-map/pull/3 ships to Webpack 1.x users without deleting `node_modules` and reinstalling.

Webpack core PR is in https://github.com/webpack/core/pull/30, this one is just for consistency and I don't care too much when this one gets released.


SpaceK33z:
Thanks!

Issue #389, opened 2016-12-09
getLocalIdent not working


VanishingDante:
I'm using css-loader@0.26.1 which is the latest version and Webpack@2.1.0-beta.25.

My css file loaders configuration is below
```
{
          test: /\.css$/,
          use: [
            { loader: 'style-loader' },
            {
              loader: 'css-loader',
              options: {
                module: true,
                // getLocalIdent: (loaderContext, localIdentName, localName, options) => `${loaderContext}-${localIdentName}-${localName}-${options}`,
                getLocalIdent: function (loaderContext, localIdentName, localName, options) {
                  console.log(arguments)
                  return 'whatever_random_class_name'
                },
                localIdentName: '[path]---[name]---[local]---[hash:base64:5]'
              }
            },
            { loader: 'postcss-loader' }
          ]
        },
```

no output in the console, the function is not even called

nowaalex:
+1

tonyhb:
+1

fengyun2:
+1

faceyspacey:
+1

Lodin:
In accordance with results of my exploration there is no function `getLocalIdent` in [query](https://github.com/webpack/css-loader/blob/master/lib/loader.js#L15) object in `loader.js`. Looks like webpack2 converts everything in the `query` field to a JSON string before using query in loader, and that strips everything except JSON-compatible data. 

Hope it will help.

soluml:
This is still an issue in webpack@2.2.0-rc.1

soluml:
Tested this with latest Webpack (as of this writing, 2.2.0-rc.2) and I'm getting the same issues.  getLocalIdent is completely ignored.

fengyun2:
css-loader@0.23.1 which is the latest version and Webpack@2.1.0-beta.28.
`localIdentName: [path]---[name]---[local]---[hash:base64:5]` 中的 `path` 没法识别

huarse:
+ 1

TanaseHagi:
```javascript
++ 1
```

nshoes:
+1

michael-ciniawsky:
Fixed by either installing webpack >= v2.2.1 or adding an `ident`  to the loader options

```js
{ 
     loader: 'css-loader`, 
     options: {
        ident: 'css',
        getLocalIdent: () => {}
        ...
     }
}
```

PR #391, opened 2016-12-14
Enable getLocalIdent in global options

hh9527:

This PR will enable global options, and thus can give a workaround to https://github.com/webpack/css-loader/issues/389 : just put getLocalIdent in global options via LoaderOptionsPlugin

Yes, in my project, I use a patch to this loader, and works fine.

Not yet. cause it still has space to improve: enable other options in global. but it should be judged by project owner.

**Summary**

See above.

https://github.com/webpack/css-loader/issues/389

No

**Other information**

This will make it possible to work with webpack2 and extract-text-webpack-plugin

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -140,7 +140,9 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
 	var forceMinimize = query.minimize;
 	var minimize = typeof forceMinimize !== "undefined" ? !!forceMinimize : options.minimize;
 
-	var customGetLocalIdent = query.getLocalIdent || getLocalIdent;
+	var customGetLocalIdent = query.getLocalIdent ||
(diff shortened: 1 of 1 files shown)


gustavohenke:
> Did you add tests for your changes?
> 
> Yes, in my project, I use a patch to this loader, and works fine.

Duh, "tests" are meant to be unit tests.

timse:
I think what you are trying to fix can be solved by adding "ident" to you loaders options. (which may or may not be automatic soon :)).

michael-ciniawsky:
> I think what you are trying to fix can be solved by adding "ident" to you loaders options.

@timse can you confirm this ? :)

michael-ciniawsky:
See reply #389, this should be fixed now :) thx

PR #393, opened 2016-12-16
Attempt creating globally unique module id's

GeorgeTaveras1231:

bugfix

Not yet, I'm interested to see the level of buy-in to these changes before spending time updating/writing new tests

Same as above. 

**Summary**

Module IDs are used to determine whether a css module has been loaded or not. However, webpack's `module.id` is only unique per webpack build. In our use case, we are trying to distribute css-modules as npm packages which are bundled separately. This causes collisions for consumers when bundling more than one dependency.

Example:

project-a's export looks like:
```javascript
[[1, '/* css for project-a omitted */', null]]
```
project-b's export looks like:
```javascript
[[1, '/* css for project-b omitted */', null]]
```

When project-c tries to import project-a and project-b, it determines that project-b is already loaded because it has the same id as project-a. [See here](https://github.com/webpack/css-loader/blob/master/lib/css-base.js#L39)

I don't think so but I don't know all of the consumption contracts.

**Other information**

We are using the query string `importLoaders=false` to bypass the css-loader on imports. This is important to us because we are distributing modules as npm packages and we think that post-bundle js files is the most scalable distribution format.

### Todo:
- [ ] Confirm whether this type of change is a good change or not.
- [ ] Update tests

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -3,6 +3,7 @@
 	Author Tobias Koppers @sokra
 */
 var path = require("path");
+var rs = require('random-strings');
 var loaderUtils = require("loader-utils");
 var processCss = require("./processCss");
 var getImportPrefix = require("./getImportPrefix");
@@ -46,7 +47,7 @@ module.exports = function(content, map) {
 			return true;
 		}).map(function(imp) {
 			if(!loaderUtils.isUrlRequest(imp.url, root)) {
-				return "exports.push([module.id, " +
+				return "exports.push(['" + rs.hex(10) +"', " +
 					JSON.stringify("@import url(" + imp.url + ");") + ", " +
 					JSON.stringify(imp.mediaQuery) + "]);";
 			} else {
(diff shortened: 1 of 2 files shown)


michael-ciniawsky:
@webpack-contrib/org-maintainers Anyone thoughts on this, personally 👎

bebraw:
👎 It would be better to handle this some other way. The suggested solution feels too low level. Ideally you could manipulate module ids elsewhere to match your requirements.

d3viant0ne:
Agreed. Global & Unique has never turned out well in my 38 years of life. Honestly, even if you could get to to work properly it would be difficult to maintain and has little benefit to the consumers of this loader. 

Thanks @GeorgeTaveras1231 but we are going to have to decline.

GeorgeTaveras1231:
Thanks for looking at this @d3viant0ne @bebraw @michael-ciniawsky.

I agree that my solution is rather hacky and there is probably a more elegant solution to @bebraw 's point.

I am surprised that my use case hasn't surfaced elsewhere though, do you guys know of anyone or any project where they are distributing css modules (using css-loader) as npm modules?

bebraw:
@GeorgeTaveras1231 [react-toolbox](https://github.com/react-toolbox/react-toolbox) is distributed using CSS Modules. Maybe examining that project could give some idea on how to handle the problem.

GeorgeTaveras1231:
Thanks, @bebraw! I will look into how they do.

PR #402, opened 2017-01-18
Initial ES6 modules support

### LINKED ISSUE #210: need to support ES6 module
I'm using webpack 2.0 beta, if I want to " import xx from 'x.css' ", then pass xx string to the style config(like styles: xx), running will go wrong cause the build will be: "styles: xx.default", however cssloader will be : "module.exports=xx", not "module['default']=xx".

simlrh:

Imports and exports changed to use the native ES6 module system.

importLoaders option changed to allow importLoaders to run after the css-loader (to allow babel to run on imported css modules).

Yes, updated tests.

Yes

**Summary**

ES6 exports work with tree shaking, which makes it possible to track which CSS classes are used and remove dead CSS (ie in a plugin that fires prior to minification).

This version requires a new style-loader to consume the exports correctly, and the same for any loader expecting output from css-loader. A PR for a compatible style-loader is being posted along with this one.

**Other information**

"@import" statements import the default object from imported css. When using "composes:" only the symbols used are imported.

CSS classes are exported separately as camel case variables. An object containing all classes is exported as the default. Classes that clash with JS reserved words are not exported separately but are included in the default. The default object has a toString function which returns the CSS itself.

In addition an object called $css is exported which contains the css source & sourcemap and a list of imported css media query information to help style-loader.

Resolves #210


lib/compile-exports.js
```diff
@@ -11,19 +11,26 @@ module.exports = function compileExports(result, importItemMatcher, camelCaseKey
...
+      symbols.push(camelCaseKey + ": " + camelCaseKey);
+    }     
+
+    res.push("export const " + camelCaseKey + " = " + valueAsString);
+    symbols.push("'" + key + "': " + camelCaseKey);
 
     return res;
-  }, []).join(",\n");
+  }, []);
+  exportJs.push("export default { " + symbols.join(", ") + " };");
```
sokra: It should not default export the key, but export each symbol separate. Best emit an

``` js
export var className = "class-name";
```

per class name.

The CSS itself should be exported via default export.

``` js
export default css;
```

sokra: This means class name always need to be converted to camelCase from dashed-case

lib/loader.js
```diff
@@ -36,35 +37,75 @@ module.exports = function(content, map) {
...
+			var importUrl = importUrlPrefix + url;
+			var importDefault = importSymbols[url].default;
+			var hash = importSymbols[url].hash;
+			var symbols = importSymbols[url].symbols.map(function(symbol) {
+				return camelCase(symbol);
+			});
+
+			var js = "import { $cssLoader as $cssLoader" + hash + " } from '" + importUrl + "';\n";
+			if (symbols.length) {
+				js = js + "import { " + symbols.join(", ") + " } from '" + importUrl + "';\n";
```
sokra: Instead of `'" + importUrl + "'` always use `JSON.stringify(importUrl)`, elsewise if breaks for chars that need to be escaped.

lib/compile-exports.js
simlrh: It does both, exports each separate symbol but also exports all symbols as properties on one default object.

This was done for compatibility with the CSS modules spec where eg `import styles from './styles.css';" is expected to give you such an object. Exporting the CSS source by default makes more sense in terms of a loader but wouldn't allow that behaviour. It also allows dashed-case inside the default object, although not in the separated symbols.

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["es2015"]
```
SpaceK33z: Is this necessary? You specify the preset in the tests specifically.

simlrh:
> It should not default export the key, but export each symbol separate. ... The CSS itself should be exported via default export.

I've tried to address this by adding a toString function which returns the CSS itself. Hopefully this is an acceptable solution as the CSS is available but it also works the way CSS Modules expect (at the expense of clashing with .toString and .to-string, if anyone is evil enough to put that in CSS).

Another tradeoff is skipping separate export of classnames which clash with JS reserved words, but these are still exported the default object.

.babelrc
simlrh: No sorry, removed

lib/css-base.js
```diff
@@ -9,7 +9,10 @@ var CssObject = function(css) {
 		enumerable: false,
 		writable: true
 	});
-	Object.assign(this, css);
+	var that = this;
+	Object.keys(css).map(function(key) {
+		that[key] = css[key];
+	})
 };
```
d3viant0ne: Where do we support `node 0.12`?

d3viant0ne:
@simlrh - Can you pick up master when you have a second, it has a bunch of applicable Travis changes.

simlrh:
Regarding using tree shaking to remove unused rules from the CSS, I've implemented this as a plugin called [webpack-dead-css](https://github.com/simlrh/webpack-dead-css), demo project [here](https://github.com/simlrh/webpack-dead-css-demo).

simlrh:
Changed importLoader behaviour to allow loaders to run before and after css-loader on imported modules. Necessary so that babel-loader can be applied to the ES6 output. Eg:

    {
        loader: 'css-loader',
        query: {
            importLoaders: {
                beforeCssLoader: [ 'postcss-loader' ],
                afterCssLoader: [{
                    loader: 'babel-loader',
                    query: {
                        presets: [ ['es2015', {modules: false}] ]
                    }
                }]
            }
        }
    }

simlrh:
Not sure now that [changing importLoader behaviour](https://github.com/webpack-contrib/css-loader/pull/402/commits/69a21cc4deacb4415518bc00aea73a55e627d12f) is the best solution.

If loaders are going to return ES2015 modules then babel-loader will need to be added to the top of the loader chain for every file type, as well as importLoaders on any imported CSS. (I had 3 babel-loader invocations in the demo project for this PR). And webpack 2 can handle ES2015 anyway, so there's no actual need to transpile in a loader any more. I think the better solution is to forget babel-loader and run babel during "optimize-chunk-assets" like UglifyJsPlugin.

See [babel-webpack-plugin](https://github.com/simlrh/babel-webpack-plugin/) for this behaviour.

simlrh:
FYI the CLA check has been broken since the repo moved organisation so I can't access it to sign again, but I did previously sign the agreement: https://github.com/webpack-contrib/css-loader/pull/402#issuecomment-273567853

michael-ciniawsky:
@simlrh This PR is definitely stalled which is shame tbh, contains a lot of the fixes we recently merged or working on. If you still have the motivation to rebase this PR and refactor all the superseded parts out feel free to do so and let my know :)

⚠️ Note that ES2015 Modules support will be part of the upcoming`webpack-defaults`upgrade alongside with some other breaking changes.

Otherwise I will close this PR and for still outstanding issues that were adressed by this PR feel free to send PR after the `webpack-defaults` upgrade, I ensure it will not get stalled again 😛

michael-ciniawsky:
Closing for now but feel free to reopen :)

PR #416, opened 2017-02-02
Fix line endings in README.md

christianvuerings:
**What kind of change does this PR introduce?**
Line endings fix for `README.md`

**Did you add tests for your changes?**
No

**If relevant, did you update the README?**
Yes

**Summary**
For more info see https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings

**Does this PR introduce a breaking change?**
No

PR #420, opened 2017-02-14
fix: do not export duplicate keys

joscha:

Bugfix, it makes sure that keys that are the same in different case styles (e.g. dashed, camelCase, dashedCamelCase) are not added to the resulting output when `?camelCase` is enabled.

E.g. a class name `.bla` would be `bla` as a default, but also `bla` in camelCase whereas `.my-bla` would yield `my-bla` and `myBla`. Currently the constructed document for the locals contains two entries, like this:
```json
{
  "bla": "_1L-rnCOXCE_7H94L5XT4uB",
  "bla": "_1L-rnCOXCE_7H94L5XT4uB",
}
```
if `camelCase` is enabled.

No, because the test helper uses JSON deserialization, which removes the dupe.

**Summary**

When using CSS modules with Typescript for example, you encounter something like this:
```
ERROR in /Users/joscha/dev/web-ui/src/ui/button/button.css.ts
(11,2): error TS2300: Duplicate identifier '"defaultFontStack"'.

ERROR in /Users/joscha/dev/web-ui/src/ui/button/button.css.ts
(13,2): error TS2300: Duplicate identifier '"button"'.
```
Also, the resulting output in the bundle is 50% bigger than it has to be if only class names are used that are `className === camelCase(className)`.

No

### DIFF
--- a/lib/compile-exports.js
+++ b/lib/compile-exports.js
@@ -14,14 +14,23 @@ module.exports = function compileExports(result, importItemMatcher, camelCaseKey
   var exportJs = Object.keys(result.exports).reduce(function(res, key) {
     var valueAsString = JSON.stringify(result.exports[key]);
     valueAsString = valueAsString.replace(result.importItemRegExpG, importItemMatcher);
-    res.push("\t" + JSON.stringify(key) + ": " + valueAsString);
-
-    if (camelCaseKeys === true) {
-      res.push("\t" + JSON.stringify(camelCase(key)) + ": " + valueAsString);
-    } else if (camelCaseKeys === 'dashes') {
-      res.push("\t" + JSON.stringify(dashesCamelCase(key)) + ": " + valueAsString);
+    function addEntry(k) {
+      res.push("\t" + JSON.stringify(k) + ": " + valueAsString);
     }
+    addEntry(key);
 
+    var targetKey;
+    if (camelCaseKeys === true) {
+      targetKey = camelCase(key);
+      if (targetKey !== key) {
+        addEntry(targetKey);
+      }
+    } else if (camelCaseKeys === 'dashes') {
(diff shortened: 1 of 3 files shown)


joscha:
can't find the place to sign the CLA

joscha:
still no email from the CLA bot :(

michael-ciniawsky:
@joscha Please close and reopen the PR once again for the CLA, it should work now :)

bebraw:
I think you have to close/open the PR again as the CLA bot was fixed about ten days ago.

joscha:
@michael-ciniawsky @bebraw done

joscha:
@d3viant0ne I tried keeping the diff as small as possible - made some changes so it is more readable now, please take another look.

joscha:
@bebraw sure, I will need to change the current tooling a tiny bit for that because right now it is all based on `JSON.parse` which removes the duplication before we can test it. If that's okay I am happy to add a test for it.

michael-ciniawsky:
@joscha From my side, just move on :)

bebraw:
@joscha Ok, cool. Safer with tests. 👍

joscha:
@bebraw @michael-ciniawsky I added two tests however, there is something really strange happening with the `README.md` - somehow the line endings get converted and any `git config core.autocrlf true`, etc. couldn't change it so the `README.md` doesn't show as changed. Do you have any idea how to fix this, as currently the diff contains the line ending changes.

joscha:
The line endings in README seem to be fixed by #416 - for now you can just squash my changes.

joscha:
I squashed all commits into one, so should be good to go now 👍

bebraw:
Nice work. I'll let either Michael or Joshua to merge and publish. 👍

Issue #280, opened 2016-05-27
Source Maps generating wrong file paths


cevou:
When creating source maps with the CSS loader the file paths that are generated are wrong and don't follow the format which is used for JavaScript files.

CSS Loader defines a sourceRoot (`webpack://`) for the source map. Additionally `webpack://` is added a second time within the `SourceMapDevToolPlugin` when the file names are generated using `ModuleFilenameHelpers`. This finally results in this file path for the css file: `'webpack:///webpack:///components/Test.css'`

See this simple Example:

```
src
-- components
-- -- Test.css
-- -- Test.js
-- app.js
webpack.config.js
```

with `Test.js`:

``` js
import React, { Component, PropTypes } from 'react';
import './Test.css';

export default class Test extends Component {
    render() {
        return (
            <div className="test"></div>
        );
    }
}
```

and `app.js`:

``` js
import React from 'react';
import { render } from 'react-dom';
import Test from './components/Test'

render(<Test />,  document.getElementById('demo'))
```

and `webpack.config.js`:

``` js
import path from 'path';
import ExtractTextPlugin from 'extract-text-webpack-plugin';

const srcPath = path.join(__dirname, 'src');
const buildPath = path.join(__dirname, 'dev');

module.exports = {
    context: srcPath,
    entry: [
        './app.js'
    ],
    output: {
        filename: '[name].bundle.js',
        chunkFilename: '[name].chunk.js',
        path: buildPath
    },
    module: {
        loaders: [
            {
                test: /\.js$/,
                exclude: /node_modules/,
                loaders: ['babel-loader']
            },
            {
                test: /\.css/,
                loader: ExtractTextPlugin.extract('style', ['css?sourceMap'])
            }
        ]
    },
    plugins: [
        new ExtractTextPlugin('[name].style.css')
    ],
    devtool: 'source-map'
};
```

If you skip into the source map creation you can see that `moduleFileNames` in `SourceMapDevToolPlugin` is correct for the JS files:

```
[
  'webpack:///./components/Test.js',
  'webpack:///./components/Test.css'
]
```

However, if you look at the `moduleFileNames` for the source map from the CSS Loader you get:

```
[
  'webpack:///webpack:///components/Test.css'
]
```

This is obviously wrong. I would expect the same path that was generated for the JS files.

While digging a little bit deeper into it I found out that the JS source maps that are processed always contain the absolute path to the JS file and an empty `sourceRoot` property.

[...]

PR #430, opened 2017-02-24
use absolute paths and remove sourceRoot when generating sourcemaps

### LINKED ISSUE #373: Problem with hard-coded sourceRoot
I'm using css-loader (with sass-loader), and I'm finding that CSS source maps are not working in Chrome or Firefox. The generated CSS is being dynamically inserted by style-loader, and in the Chrome developer tools Styles panel, all of the source links look like this:

    blob:http://localhost:8080/022723dd-255f-4c67-8da3-486c82cba226

When clicked, these links take me to the generated CSS and not the original source SCSS file.

It seems as if this problem is related to the hard-coded sourceRoot being set by css-loader (https://github.com/webpack/css-loader/blob/master/lib/loader.js#L106):

    map.sourceRoot = "webpack://";

Chrome doesn't seem to know how to interpret this and find the source files, and I can't see any way to override this sourceRoot, so I tried temporarily changing that line of code to this:

    map.sourceRoot = "http://localhost:8080";

...and the source maps started working. Am I missing something or is this a problem with css-loader?

brokenmass:
bugfix

Yes

Not relevant

**Summary**
Simplifies and correct sourcemaps generation by returning absolute sources paths and removing sourceRoot.
Fix #373 and #280

No

**Other information**

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -2,7 +2,6 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-var path = require("path");
 var loaderUtils = require("loader-utils");
 var processCss = require("./processCss");
 var getImportPrefix = require("./getImportPrefix");
@@ -97,13 +96,9 @@ module.exports = function(content, map) {
 			map = result.map;
 			if(map.sources) {
 				map.sources = map.sources.map(function(source) {
-					source = source.split("!").pop();
-					var p = path.relative(query.context || this.options.context, source).replace(/\\/g, "/");
-					if(p.indexOf("../") !== 0)
-						p = "./" + p;
-					return "/" + p;
+					return source.split("!").pop();
 				}, this);
-				map.sourceRoot = "webpack://";
+				map.sourceRoot = "";
 			}
 			map.file = map.file.split("!").pop();
 			map = JSON.stringify(map);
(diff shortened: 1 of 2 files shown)


brokenmass:
any updates ?

d3viant0ne:
@brokenmass for reference ping  `@webpack-contrib/org-maintainers` as it's easy to miss comments in the news feed and watching 43 repositories is a nightmare :)

jasisk:
This change appears to introduce some complexity for using, e.g., chrome dev tools.

Because `css-loader` is no longer following the same convention as webpack's source mapping, you end up having to add multiple workspace mappings. I presently store my css alongside the component it corresponds to. I could not figure out the exact incantation of mappings to get everything to work.

Additionally, without a protocol, Chrome treats the path as an absolute path _after_ the host. So, on a Mac, it resolves to `http://something.website:port/Users/myUserName/...`.

d3viant0ne:
@jasisk - Can you get this into a new issue so we are sure this can be addressed one way or another

michael-ciniawsky:
Yep please open a new issues this is a very inconvenient regression to said it politely, we need to investigate in a clean way 😛

davicho:
I find paths absolute slightly inconvenient. As I inspect my css in the dev tools.... the path can be so long it truncates the actual file name - making have to leave my work are to hover of the path and get the tooltip to see the actual file name I need to work on. (Unless, there is an option to undo this but I have looked and have not seen such). Ty.
truncated-path

PR #445, opened 2017-03-09
feat: allow removal of original class name

### LINKED ISSUE #368: camelCase doesn't delete original names
`camelCase` option adds new props to style object, not replaces original ones. 

It causes dirt when used with https://github.com/javivelasco/react-css-themr.

For instance, if I have `.back-button {...}` class, css-loader with `camelCase` returns 
```
{
   back-button: '...',
   backButton: '...'
}
```
And if that style object is passed to react-css-themr as `theme={styles} themeNamespace='back'`, it transforms it to 
```
{
   -button: '...',
   button: '...'
}
```
It doesn't make any sense. Is there any reasons why original class name is saved by css-loader?

### LINKED ISSUE #440: Add support for removing original classnames when using camelCase=true/'dashes'
Feature Request

CSS with both original and camelized classnames are retained in the final output.

Would it be possible to introduce yet another flag which removes the uncamelized versions from the final output? I'm asking because of a comment on one of the issues: clinyong/vscode-css-modules#3 (comment)

Basically, by retaining both the camelized and original versions of a classname, the bundle size increases, sometimes by a lot for larger projects. If all developers are in with the syntax, perhaps it'd not be as problematic to remove the original classnames? Two extra options can be added to the camelCase options: camelCase: only and camelCase: dashesOnly which cause the plugin to not output the original classes. What do you think of the idea?

I'd be happy to make a PR if the team decides it's worthwhile to add to the loader.

joscha:
feature/performance

yes

yes

**Summary**
* fixes #440 
* fixes #368 

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,6 +421,15 @@ By default, the exported JSON keys mirror the class names. If you want to cameli
 import { className } from 'file.css';
 ```
 
+#### Possible options
+
+|Option|Behaviour|
+|:----:|:--------|
+|**`true`**|Class names will be camelized|
+|**`'dashes'`**|Only dashes in class names will be camelized|
+|**`'only'`** |Class names will be camelized, the original class name will be removed from the locals|
+|**`'dashesOnly'`**|Dashes in class names will be camelized, the original class name will be removed from the locals|
+
 <h2 align="center">Maintainers</h2>
 
 <table>
--- a/README.md
+++ b/README.md
@@ -398,6 +398,15 @@ You can also disable or enforce minification with the `minimize` query parameter
 
 By default, the exported JSON keys mirror the class names. If you want to camelize class names (useful in JS), pass the query parameter `camelCase` to css-loader.
 
+#### Possible Options
+
+|Option|Behaviour|
+|:----:|:--------|
+|**`true`**|Class names will be camelized|
+|**`'dashes'`**|Only dashes in class names will be camelized|
+|**`'only'`** |Class names will be camelized, the original class name will be removed from the locals|
+|**`'dashesOnly'`**|Dashes in class names will be camelized, the original class name will be removed from the locals|
+
 **webpack.config.js**
 ```js
 {
@@ -420,15 +429,6 @@ By default, the exported JSON keys mirror the class names. If you want to cameli
 ```js
 import { className } from 'file.css';
 ```
-
-#### Possible options
-
-|Option|Behaviour|
-|:----:|:--------|
-|**`true`**|Class names will be camelized|
-|**`'dashes'`**|Only dashes in class names will be camelized|
-|**`'only'`** |Class names will be camelized, the original class name will be removed from the locals|
-|**`'dashesOnly'`**|Dashes in class names will be camelized, the original class name will be removed from the locals|
 
 <h2 align="center">Maintainers</h2>
 
--- a/README.md
+++ b/README.md
@@ -400,7 +400,7 @@ By default, the exported JSON keys mirror the class names. If you want to cameli
 
 #### Possible Options
 
-|Option|Behaviour|
+|Option|Description|
 |:----:|:--------|
 |**`true`**|Class names will be camelized|
 |**`'dashes'`**|Only dashes in class names will be camelized|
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -394,6 +394,15 @@ By default, the exported JSON keys mirror the class names. If you want to cameli
 import { className } from 'file.css';
 ```
 
+#### Possible options
```
michael-ciniawsky:
```diff
- #### Possible Options 
```

README.md
```diff
@@ -394,6 +394,15 @@ By default, the exported JSON keys mirror the class names. If you want to cameli
...
+
+|Option|Behaviour|
```
michael-ciniawsky: Move the table directly under the `camelCase` header please

`Behaviour` => `Description` for consistency

package.json
```diff
@@ -1,12 +1,16 @@
 {
   "name": "css-loader",
-  "version": "0.26.2",
+  "version": "0.26.4",
```
ekulabuhov: This is done automatically. No need to change this file.

michael-ciniawsky: @ekulabuhov Good catch 💯 I totally missed it. 

In a short while, if seen this a few times now, maybe we should update the `PR_TEMPLATE` to explicitly mention that `npm version` is always handled/done by us? cc @bebraw @ d3viant0ne

bebraw: Yeah, open an issue at **webpack-defaults** so it gets sorted once and for all.

lib/loader.js
```diff
@@ -97,13 +96,9 @@ module.exports = function(content, map) {
 			map = result.map;
 			if(map.sources) {
 				map.sources = map.sources.map(function(source) {
-					source = source.split("!").pop();
-					var p = path.relative(query.context || this.options.context, source).replace(/\\/g, "/");
-					if(p.indexOf("../") !== 0)
-						p = "./" + p;
-					return "/" + p;
+					return source.split("!").pop();
```
ekulabuhov: Is this related?

ekulabuhov:
It seems like this PR needs a rebase.

d3viant0ne:
And please note, it need a `rebase` not a merge update or conflict resolution via github.

If you aren't comfortable with the rebase comment, give me write access to your fork and I can either do it or walk you through the process

README.md
joscha: done, 8ce02a0

README.md
joscha: done, 9574dc0

joscha:
> And please note, it need a rebase not a merge update or conflict resolution via github.

rebased onto `master`, please have another look @d3viant0ne @ekulabuhov

ekulabuhov:
👍 Great work! 

Would it make more sense to opt in into generating original class names? Something like:
`{ camelCase: 'withOriginal' }`. I don't think many people depended or actually knew about this buggy behaviour. And in version 1.0 we're going to break it once again for people that switch to `only` or `dashesOnly` right now. Any opinions?

joscha:
> I don't think many people depended or actually knew about this buggy behaviour.

I know at least two companies that depend on this "feature" 💃 💃 💃  :)

@ekulabuhov I'd really like to get this fix into pre-1.0.0 and the only way to not break existing code is to make the feature opt-out instead of opt-in. I am happy to pick the PR to any 1.0.0 feature branch and make it opt-in to keep the original name, if there is one?

d3viant0ne:
@joscha - Yeah, this should go in before the next major. Backwards compatible and beneficial to both major versions.

joscha:
> @joscha - Yeah, this should go in before the next major. Backwards compatible and beneficial to both major versions.

@d3viant0ne agreed, so I think the current PR is gtg 🏃

d3viant0ne:
I'll merge this as soon as I finish up with a few other reviews.

joscha:
I made a mistake on this - due to the change from #420 simple classes or ones that were already in camelCase are not exported any more because they are thought to be already on the object. This only happens when one of the new options is enabled. Will write a fix and send in a PR.

joscha:
ever so sorry, the test cases were not wide enough and it wasn't picked up in the review - the fix is here: #448

PR #448, opened 2017-03-10
fix: incorrect handling of non-mutated class names

joscha:
bugfix

yes

n/a

**Summary**

When converting a class name without dashes, the newly introduced `'only'` and `'dashesOnly'` settings for the `camelCase` option would only ever add the class name to locals if it was different to the standard class name. Any class w/o dash would therefore not be added to the locals at all.

No

**Other information**
Introduced in #445 - released in `0.27.0`

### DIFF
--- a/lib/compile-exports.js
+++ b/lib/compile-exports.js
@@ -17,21 +17,32 @@ module.exports = function compileExports(result, importItemMatcher, camelCaseKey
     function addEntry(k) {
       res.push("\t" + JSON.stringify(k) + ": " + valueAsString);
     }
-    if (camelCaseKeys !== 'only' && camelCaseKeys !== 'dashesOnly') {
-      addEntry(key);
-    }
 
     var targetKey;
-    if (camelCaseKeys === true || camelCaseKeys === 'only') {
-      targetKey = camelCase(key);
-      if (targetKey !== key) {
-        addEntry(targetKey);
-      }
-    } else if (camelCaseKeys === 'dashes' || camelCaseKeys === 'dashesOnly') {
-      targetKey = dashesCamelCase(key);
-      if (targetKey !== key) {
-        addEntry(targetKey);
-      }
+    switch(camelCaseKeys) {
+      case true:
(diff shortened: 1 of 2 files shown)


lib/compile-exports.js
```diff
@@ -17,21 +17,31 @@ module.exports = function compileExports(result, importItemMatcher, camelCaseKey
...
       res.push("\t" + JSON.stringify(k) + ": " + valueAsString);
     }
-    if (camelCaseKeys !== 'only' && camelCaseKeys !== 'dashesOnly') {
-      addEntry(key);
-    }
 
-    var targetKey;
-    if (camelCaseKeys === true || camelCaseKeys === 'only') {
-      targetKey = camelCase(key);
-      if (targetKey !== key) {
```
joscha: this line was the culprit, as it would prevent the camelcased key to be added to the locals only if it was different from the default key. However with the `only` and `dashesOnly` options the default key would not have been added -> empty locals.

joscha:
cc @d3viant0ne I believe the release of `0.27.0` is not that old, so it would probably good to have a look at this before the issues trickle in.

lib/compile-exports.js
```diff
@@ -17,21 +17,32 @@ module.exports = function compileExports(result, importItemMatcher, camelCaseKey
...
       res.push("\t" + JSON.stringify(k) + ": " + valueAsString);
     }
-    if (camelCaseKeys !== 'only' && camelCaseKeys !== 'dashesOnly') {
-      addEntry(key);
-    }
 
     var targetKey;
-    if (camelCaseKeys === true || camelCaseKeys === 'only') {
-      targetKey = camelCase(key);
-      if (targetKey !== key) {
```
joscha: this line was the culprit, as it would prevent the camelcased key to be added to the locals only if it was different from the default key. However with the only and dashesOnly options the default key would not have been added -> empty locals.

joscha:
Thank you @bebraw! And sorry about the hassle.

PR #453, opened 2017-03-13
Remove extra source map file + write source maps only if they are enabled

bebraw:

Refactoring.

Old tests should work (had to tweak one).

**Summary**

This gets rid of the older hackfix.


bebraw:
Published in `0.27.3`. Thanks for the confirmation.

fklingler:
Ok I've got some bad news... I just realized it doesn't work when the sourceMap option is present.

I have to set
```
  node: {
    Buffer: true
  }
```

in the webpack options for this to be working.

michael-ciniawsky:
@fklingler That's confusing in #454 the opposite seems to be the case :D, can you show the bundle output and maybe make a small test repo for reproduction, there is a small bug somewhere. Please move the discussion to #454 aswell to keep it at one place 😛

PR #465, opened 2017-03-23
added localByDefault option to allow you to override default behavior…

hulkish:
Feature

Yes.

Yes.

**Summary**
The default behavior when using `modules: true` uses `:local` mode by default. This option allows you to override that behavior. This is useful for porting projects which weren't previously using CSS Modules to start using it.

No.

**Other information**
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ To be compatible with existing css files (if not in CSS Module mode):
 |:--:|:-----:|:----------|
 |**`root`**|`/`|Path to resolve URLs, URLs starting with `/` will not be translated|
 |**`modules`**|`false`|Enable/Disable CSS Modules|
+|**`localByDefault`**|`true`|treat unmarked selectors as :local|
 |**`import`** |`true`| Enable/Disable @import handling|
 |**`url`**|`true`| Enable/Disable `url()` handling|
 |**`minimize`**|`false`|Enable/Disable minification|
@@ -232,6 +233,12 @@ Note: For prerendering with extract-text-webpack-plugin you should use `css-load
 The query parameter `modules` enables the **CSS Modules** spec.
 
 This enables local scoped CSS by default. (You can switch it off with `:global(...)` or `:global` for selectors and/or rules.)
+
+### :local by default
+
+The query parameter `localByDefault` allows you to override the _"locals by default when using `modules: true`"_ behavior'.
+
+For selectors that have not been marked using :local or :global - this will allow you to decide which should be default. This will help for projects that are being ported from non css modules usage to using css modules.
 
 ### CSS Composing
 
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -15,6 +15,9 @@ module.exports = function(content, map) {
 	var query = loaderUtils.getOptions(this) || {};
 	var root = query.root;
 	var moduleMode = query.modules || query.module;
+  var localByDefault = typeof query.localByDefault !== 'undefined'
+    ? query.localByDefault
+    : true;
 	var camelCaseKeys = query.camelCase || query.camelcase;
 	var resolve = createResolver(query.alias);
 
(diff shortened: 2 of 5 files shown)


michael-ciniawsky:
@hulkish  https://github.com/webpack-contrib/css-loader#css-modules ? This should be the current behaviour already, can you elaborate please ? 😛

hulkish:
@michael-ciniawsky It is default behavior *when you are not using css modules  (`modules: true`) option. However, when you use `modules: true` - this behavior defaults to using `:local` instead of `:global:.

TrySound:
@hulkish And you did an option to override `modules` option?

hulkish:
@TrySound No, this option will solely override the `local by default when modules: true` behavior. You can see this change here: https://github.com/webpack-contrib/css-loader/pull/465/files#diff-7c7d4918b118d27d18538f53b8155e88

michael-ciniawsky:
@hulkish Can you please show code before/after because I thought the same as @TrySound tbh 😛

hulkish:
@michael-ciniawsky https://github.com/webpack-contrib/css-loader/pull/465/files#diff-7c7d4918b118d27d18538f53b8155e88

TrySound:
@hulkish But this is what modules option does, isn't it? I enables local by default. With modules false it's not default but available.

hulkish:
@TrySound Yes, you are correct - I suppose the confusion is originating from the name of the option itself `localByDefault` - per my use case, I am actually setting it to `false` to get the outcome I am after.

Perhaps renaming it to `globalByDefault` would make more sense?

michael-ciniawsky:
@TrySound Yep, that should be the case 😉

michael-ciniawsky:
@hulkish But that's `{ modules: false }` then 😛 . Use `:local(.className) { ... }` to make portions local

michael-ciniawsky:
@hulkish If I missunderstand something here, please give an example how it will look like before/after

TrySound:
:global is default without options :)

hulkish:
@michael-ciniawsky wait...but solely setting localIdentName to true doesn't really enable all the css modules features, though?

TrySound:
localIdentName customizes local output

hulkish:
I want css modules to  *only* localize class names which I've selectively designated via `:local .myclass {}`

TrySound:
Then it should work without options.

michael-ciniawsky:
@hulkish That shold work on global mode

hulkish:
I guess I'm confused by the notion... So you're saying that I can use css modules without setting `modules: true`, and only defining a `localIdentName`?

Here's the scenario I'm in..:

- Porting an existing app which relies heavily on globals. (bootstrap, etc etc)
- Due to the sheer amount of work it's going to take to port all styles to use css modules is considerably large. 
- Would like to progressively port the application to using css modules without breaking everything that hasn't yet been ported to use css modules.

michael-ciniawsky:
@hulkish Vendors (e.g Bootstrap) will need to stay global and it is better to reference them as static assets via `<link href="">` (CDN) (faster build) in e.g `index.html`. Your current App will need bigger refactors anyways, when it should use a component based architecture 😛 

#### app/components/component1/*.{css, js}

**index.css**
```css
.btn {
   background-color: red;
}
```
**index.js**
```
import styles from './style.css'

function render () {
 // styles.btn === "BHML95n" 
 return `<div class="container"><button class=${styles.btn}></button></div>`
}
```

#### app/components/component2/*.{css, js}

**index.css**
```
.btn {
   background-color: green;
}
```
**index.js**
```
import styles from './index.css'

function render () {
  // styles.btn === "KhJZrL85n" 
 return `<div class="container"><button class=${styles.btn}></button></div>`
}
```

hulkish:
@michael-ciniawsky I agree they would be better moved to a CDN - but the reality is that doing the refactor in one single motion introduces more risk than we can afford. 

Also, this line here leads me to to believe that there's more to it than just simply defining a `localIdentName`.. does `:local` work when `modules` is not true?

michael-ciniawsky:
Yep relevant lines are below, but I can't speak for your particular code base and which caveats this may have or not 😛 

https://github.com/webpack-contrib/css-loader/blob/master/lib/processCss.js#L138 (Query)
https://github.com/webpack-contrib/css-loader/blob/master/lib/processCss.js#L154 (Mode)
https://github.com/webpack-contrib/css-loader/blob/master/lib/processCss.js#L171 (Scope [LocalIdent])

I close this PR since it duplicates the `module` option thx  🙃

PR #473, opened 2017-03-31
Update README.md

briandipalma:
Docs

No

Yes

**Summary**
Previous wording was awkward, not sure the suggested one is perfect so you can reject this.

No

**Other information**

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,8 +205,9 @@ You can configure the generated ident with the `localIdentName` query parameter 
 }
 ```
 
-You can also specify the absolute path to your custom `getLocalIdent` function to generate classname based on a different schema. Note that this requires `webpack >= v2.x.` since to be able to pass function in. For example:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -205,8 +205,9 @@ You can configure the generated ident with the `localIdentName` query parameter
 }
 ```
 
-You can also specify the absolute path to your custom `getLocalIdent` function to generate classname based on a different schema. Note that this requires `webpack >= v2.x.` since to be able to pass function in. For example:
+You can also specify the absolute path to your custom `getLocalIdent` function to generate classname based on a different schema. This requires `webpack >= v2.x.` (it supports functions in the `options` object). For example:
```
michael-ciniawsky: `webpack >= 2.x` => `webpack >= 2.2.1` bc before that version using functions in options would require to specify an `{ loader: 'css-loader', options: { ident: 'css-loader', ...options } }`  to work properly

briandipalma:
Review comment addressed.

Issue #165, opened 2015-10-16
css-loader don't load a file


franckmaurin:
Hello,

I've a strange bug when I compile with webpack my react project (using css-loader):
```
ERROR in ./~/css-loader!./~/cssnext-loader!./src/components/Work/Single/Nav/styles.css
TypeError: Cannot read property 'map' of undefined

ERROR in ./~/css-loader!./~/cssnext-loader!./src/components/Work/Single/Nav/styles.css
Module build failed: TypeError: Cannot read property 'toString' of undefined
    at new Input (/Users/franckmaurin/Dropbox/Pro/Server/dev/franckmaurin.lh/node_modules/css-loader/node_modules/postcss/lib/input.js:31:23)
    at parse (/Users/franckmaurin/Dropbox/Pro/Server/dev/franckmaurin.lh/node_modules/css-loader/node_modules/postcss/lib/parse.js:21:17)
    at new LazyResult (/Users/franckmaurin/Dropbox/Pro/Server/dev/franckmaurin.lh/node_modules/css-loader/node_modules/postcss/lib/lazy-result.js:57:24)
    at Processor.process (/Users/franckmaurin/Dropbox/Pro/Server/dev/franckmaurin.lh/node_modules/css-loader/node_modules/postcss/lib/processor.js:36:16)
    at processCss (/Users/franckmaurin/Dropbox/Pro/Server/dev/franckmaurin.lh/node_modules/css-loader/lib/processCss.js:165:11)
    at Object.module.exports (/Users/franckmaurin/Dropbox/Pro/Server/dev/franckmaurin.lh/node_modules/css-loader/lib/loader.js:22:2)
 @ ./src/components/Work/Single/Nav/styles.css 4:14-143

ERROR in ./src/components/Work/Single/Nav/styles.css
Module build failed: ModuleError: TypeError: Cannot read property 'map' of undefined
    at DependenciesBlock.<anonymous> (/Users/franckmaurin/Dropbox/Pro/Server/dev/franckmaurin.lh/node_modules/webpack/node_modules/webpack-core/lib/NormalModuleMixin.js:79:21)
    at Object.cssnextLoader (/Users/franckmaurin/Dropbox/Pro/Server/dev/franckmaurin.lh/node_modules/cssnext-loader/index.js:36:10)
```

If I console.log the content in the /css-loader/lib/loader.js, I get "undefined" for one css file (Work/Single/Nav/styles.css).

The css file in question has this following code :
```
.workSingle-nav {
[...]

PR #478, opened 2017-04-08
perf: generate source maps only when explicitly set

evilebottnawi:
bugfix

not required

not required

**Summary**

https://github.com/webpack-contrib/sass-loader/issues/296#issuecomment-288464857

No

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -16,10 +16,16 @@ module.exports = function(content, map) {
 	var root = query.root;
 	var moduleMode = query.modules || query.module;
 	var camelCaseKeys = query.camelCase || query.camelcase;
+	var sourceMap = query.sourceMap || false;
 	var resolve = createResolver(query.alias);
 
-	if(map !== null && typeof map !== "string") {
-		map = JSON.stringify(map);
+	if(sourceMap) {
+		if (map && typeof map !== "string") {
+			map = JSON.stringify(map);
+		}
+	} else {
+		// Some loaders (example `"postcss-loader": "1.x.x"`) always generates source map, we should remove it
+		map = null;
 	}
 
 	processCss(content, map, {
@@ -28,7 +34,8 @@ module.exports = function(content, map) {
 		to: loaderUtils.getCurrentRequest(this),
 		query: query,
 		minimize: this.minimize,
-		loaderContext: this
+		loaderContext: this,
+		sourceMap: sourceMap
 	}, function(err, result) {
 		if(err) return callback(err);
 
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -200,12 +200,12 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
 		// we need a prefix to avoid path rewriting of PostCSS
 		from: "/css-loader!" + options.from,
 		to: options.to,
-		map: {
+		map: options.sourceMap ? {
 			prev: inputMap,
 			sourcesContent: true,
 			inline: false,
 			annotation: false
-		}
+		} : null
 	}).then(function(result) {
 		callback(null, {
 			source: result.css,
--- a/test/sourceMapTest.js
+++ b/test/sourceMapTest.js
@@ -10,6 +10,24 @@ describe("source maps", function() {
 	testWithMap("falsy: undefined map doesn't cause an error", ".class { a: b c d; }", undefined, [
 		[1, ".class { a: b c d; }", ""]
 	]);
+    testWithMap("should don't generate sourceMap when `sourceMap: false` and map exist",
+		".class { a: b c d; }",
+        {
+            file: 'test.css',
+            mappings: 'AAAA,SAAS,SAAS,EAAE',
+            names: [],
+            sourceRoot: '',
+            sources: [ '/folder/test.css' ],
+            sourcesContent: [ '.class { a: b c d; }' ],
+            version: 3
+        },
+		[
+			[1, ".class { a: b c d; }", ""]
+		],
+		{
+            query: "?sourceMap=false"
+		}
+	);
 	testMap("generate sourceMap (1 loader)", ".class { a: b c d; }", undefined, {
 		loaders: [{request: "/path/css-loader"}],
 		options: { context: "/" },
@@ -94,5 +112,14 @@ describe("source maps", function() {
 			sourcesContent: [ '.class { a: b c d; }' ],
 			version: 3
 		}]
+	]);
+	testMap("don't generate sourceMap (1 loader)", ".class { a: b c d; }", undefined, {
+		loaders: [{request: "/path/css-loader"}],
+		options: { context: "/" },
+		resource: "/folder/test.css",
+		request: "/path/css-loader!/folder/test.css",
+		query: "?sourceMap=false"
+	}, [
+		[1, ".class { a: b c d; }", ""]
 	]);
 });


bebraw:
@ekulabuhov Can you fix Node 7? It's choking on `error  Mixed spaces and tabs  no-mixed-spaces-and-tabs` at eslint.

lib/loader.js
```diff
@@ -16,13 +16,17 @@ module.exports = function(content, map) {
...
 
-	if(map !== null && typeof map !== "string") {
+	if(sourceMap && map !== null && typeof map !== "string") {
 		map = JSON.stringify(map);
+	} else {
+		map = null;
 	}
 
 	processCss(content, map, {
+        moduleMode: moduleMode,
```
bebraw: Why to pass `moduleMode` here? I don't see related code in the new tests.

lib/loader.js
```diff
@@ -16,13 +16,17 @@ module.exports = function(content, map) {
...
 	var moduleMode = query.modules || query.module;
 	var camelCaseKeys = query.camelCase || query.camelcase;
+	var sourceMap = query.sourceMap || false;
 	var resolve = createResolver(query.alias);
 
-	if(map !== null && typeof map !== "string") {
+	if(sourceMap && map !== null && typeof map !== "string") {
 		map = JSON.stringify(map);
+	} else {
+		map = null;
```
bebraw: Will it fail without `map = null`?

evilebottnawi: @bebraw yep, example `postcss-loader` now always generate source map and send their, we should remove their

bebraw: Ok, add a comment and a test so we don't lose the behavior in the future.

lib/loader.js
evilebottnawi: @bebraw stupid error, sorry

lib/loader.js
evilebottnawi: @bebraw https://github.com/webpack-contrib/css-loader/pull/478/commits/caf9873b0e1611d01ec894aeb52bdc7768413af4#diff-bbbe73d87e5520ca3a3e6b634cfae1d9R13 I have already added a test, what else is needed?

evilebottnawi: Comment added

michael-ciniawsky:
@d3viant0ne 👍

lib/loader.js
```diff
@@ -16,10 +16,14 @@ module.exports = function(content, map) {
 	var root = query.root;
 	var moduleMode = query.modules || query.module;
 	var camelCaseKeys = query.camelCase || query.camelcase;
+	var sourceMap = query.sourceMap || false;
 	var resolve = createResolver(query.alias);
 
-	if(map !== null && typeof map !== "string") {
+	if(sourceMap && map !== null && typeof map !== "string") {
```
jhnns: I think this condition is not correct. If `sourceMap` is true, but the given source map is already a string (which is uncommon in the loader eco-system but can happen as many tools like node-sass return a source map string), it will set `map = null`.

These checks need to be separated I guess.

lib/loader.js
```diff
@@ -16,10 +16,14 @@ module.exports = function(content, map) {
...
 		map = JSON.stringify(map);
+	} else {
+        // Some loaders (example `"postcss-loader": "1.x.x"`) always generates source map, we should remove it
+		map = null;
```
jhnns: Good catch. I didn't know that.

lib/loader.js
```diff
@@ -28,7 +32,8 @@ module.exports = function(content, map) {
 		to: loaderUtils.getCurrentRequest(this),
 		query: query,
 		minimize: this.minimize,
-		loaderContext: this
+		loaderContext: this,
+        sourceMap: sourceMap
```
jhnns: Is there an indentation error? 😁

evilebottnawi: @jhnns yep, we use tabs, i don't use tabs :smile:

lib/loader.js
evilebottnawi: @jhnns If it were more accurate, almost all loaders using postcss generate maps even if they are not requested :sob:

lib/loader.js
evilebottnawi: @jhnns less condition - fast work? How best this do?

lib/loader.js
michael-ciniawsky: We should fix it in `postcss-loader` aswell , I spoke with andrey a while ago about this, but this was added via PR for Angular* folks, if we have clarification of souremaps 'do && don'ts'  I will make sure it works across the whole CSS Pipeline 😛 . 

* They wanted inline sourcemaps because of the angular component system (`{String}` only)

There is a PR open at `style-loader`  [#165](https://github.com/webpack-contrib/style-loader/pull/165) aswell, which would enable inline sourcemaps support in `style-loader`, the question is if that should be working in general or not and imho whatever case is the better should be 'more strictly ' followed to avoid confusion about it. I closed a few PR's in regard of sourcrmaps already, there is definitely room for UX improvements :)

lib/loader.js
michael-ciniawsky: @evilebottnawi Feel free to convert tabs to spaces if you are in the mood, will be done via the upcoming `webpack-defaults` upgrade anyways :)

lib/loader.js
evilebottnawi: @michael-ciniawsky already reported https://github.com/postcss/postcss-loader/issues/195

evilebottnawi:
@michael-ciniawsky Can we merge this PR before `webpack-defaults` and release patch? It is really speed up build on watch if yo don't want souce maps?

michael-ciniawsky:
@evilebottnawi Yeah sure 😛, I only mentioned it in case you quickly want to go through the files and convert tabs => spaces, not important for this PR feel free to ignore

lib/loader.js
michael-ciniawsky: yep 😉 I know, but we need to make clear how we want to handle sourcemaps in the 'overall' picture and then fix it in the loaders e.g

- `{ sourceMap: false }` => no sourcemaps (absolutely, if one loader in the chain is `false`, discard them enterily)
- `{ sourceMap: true }` => as `map` (external)
- `{ sourceMap: 'inline' }` => in the CSS (`{String}`), not in favour of personally, but maybe some folks e.g Angular need it :stuck_out_tongue:

evilebottnawi: @michael-ciniawsky looks good, but I would like to know what `webpack` members think about this.
/cc @sokra

d3viant0ne: @evilebottnawi - As a general rule of thumb, you aren't going to get a hold of Tobias via Github notifications and more so within the `webpack-contrib` organization.

In the contrib org, the org-maintainers or admins will contact Tobias ( or another core member ) only when a problem or change requires a deeper level of knowledge than we collectively have. The `webpack` members are focusing on webpack core development and maintenance, we are focused on everything under contrib.

Within the scope of this pull request, @jhnns has done most of the recent performance testing and it's his input that is really required in this case.

d3viant0ne:
And yes, @evilebottnawi we will get this out prior to the defaults upgrade so it is of benefit to the current major version range.

lib/loader.js
evilebottnawi: @d3viant0ne thanks for clarify :smile:

lib/loader.js
evilebottnawi: /cc @jhnns

lib/loader.js
michael-ciniawsky: @evilebottnawi Forget abut the `{sourceMap: 'inline' }` stuff for now. Just `{ sourceMap. true }` => Generate Sourcemap && `{ sourceMap: false }` discard `map` it enterily

```
{
  file: '...' // ignored by webpack according to @sokra
  mappings: '' // update 
  sourceRoot: '' // needs to be empty according to @sokra
  sourceContent: '' // add if undefined (no previous loader) or leave 
  sources: [] // update (absolute path)
}
```

evilebottnawi: @michael-ciniawsky As I understood correctly, we should generally drop `sourceMap` option? and see on `map` from other loaders?

michael-ciniawsky: @evilebottnawi You mean @jhnns comment [here](https://github.com/webpack-contrib/css-loader/pull/478#discussion_r111349205) ?

evilebottnawi: @michael-ciniawsky not :smile: https://github.com/webpack-contrib/css-loader/pull/478#discussion_r111926378

michael-ciniawsky: ahh nope that's just about how the 'important' parts of `map` should look like to work with webpack 😛 . We need to ensure not to alter props on `map` which eventually cause issues with how webpack joins them together afterwards etc

michael-ciniawsky:
Guys, what's the status here, anything outstanding ? 😛

michael-ciniawsky:
@evilebottnawi status :) ?

evilebottnawi:
@michael-ciniawsky wait answer https://github.com/webpack-contrib/css-loader/pull/478#discussion_r111349205

lib/loader.js
michael-ciniawsky: status ? 😛

jhnns: @evilebottnawi What do you mean by less condition - fast work? 😁

jhnns: I think, the condition should look like:

```javascript
if(sourceMap) {
    if(map && typeof map !== "string") {
        map = JSON.stringify(map);
    }
} else {
    map = null;
}
```

Of course, you can avoid the nesting, but writing it like this makes it pretty clear ... at least to me 😁

jhnns:
Sorry for being unresponsive... I'm flooded with notifications 😞

Feel free to merge this if you think that my comments have been addressed.

PR #483, opened 2017-04-11
Fix package.json's engines.node field

### LINKED ISSUE #482: package.json engines field value is not working as intended

A small issue.

`package.json`'s engines `node` field is currently defined as `">=0.12.0 || >=4.3.0 <5.0.0 || >=5.10"`.

**If the current behavior is a bug, please provide the steps to reproduce.**

Try to install the package under io.js. It won't complain about an unsupported Node version being used.

The range seems to suggest only new enough versions of Node 0.12, 4 & >=5 are supported. Therefore, I'd expect the range to be something like `">=0.12 <0.13 || >=4.3 <5 || >=5.10"`.

It's clear the intent of this version range was different to what's specified.

N/A

mgol:

Bugfix.

N/A

N/A

**Summary**

`package.json`'s engines `node` field used to be defined as `">=0.12.0 || >=4.3.0 <5.0.0 || >=5.10"`. This range was erroneously equivalent to just `>=0.12`. Now only version betwee `4.3 & 5` or `>=5.10` are picked up by the range.

Fixes #482

That depends. Users of Node 0.12 & npm 2 combo will get an error during the installation of this package after this PR is merged. Whether that's considered a breaking change I'll leave for maintainers to decide.

**Other information**

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "author": "Tobias Koppers @sokra",
   "description": "css loader module for webpack",
   "engines": {
-    "node": ">=0.12.0 || >=4.3.0 <5.0.0 || >=5.10"
+    "node": ">=4.3 <5 || >=5.10"
   },
   "files": [
     "index.js",


michael-ciniawsky:
Review please, since this isn't really semver compat even it's just an tiny metadata change 😛

Issue #484, opened 2017-04-12
Source Map + Inline Images = Invalid CSS


IdanCo:
**BUG**: When processing embed images in css with the `SourceMap` flag set to `true`, invalid css is outputted. Originated from bootstrap 4.

I'm trying to compile the following css:
```css
.element {
    background-image: url("data:image/svg+xml;charset=utf8,%3Csvg viewBox='0 0 32 32' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath stroke='rgba(0, 0, 0, 0.5)' stroke-width='2' stroke-linecap='round' stroke-miterlimit='10' d='M4 8h24M4 16h24M4 24h24'/%3E%3C/svg%3E");
}
```
code is based [on this](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L617)

When the `SourceMap` flag is off, all is fine and well. But for some reason, when the `SourceMap` flag is on, it is compiled to this **invalid css** -
```css
.element {
  background-image: url("http://localhost:8080/\"data:image/svg+xml;charset=utf8,%3Csvg viewBox='0 0 32 32' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath stroke='rgba(0, 0, 0, 0.5")' stroke-width='2' stroke-linecap='round' stroke-miterlimit='10' d='M4 8h24M4 16h24M4 24h24'/%3E%3C/svg%3E"); }
```
notice that `http://localhost:8080/` suddenly popped at the beginning, which makes the css invalid.

Turning off URL handling doesn't work as mentioned here - https://github.com/webpack-contrib/css-loader/issues/44#issuecomment-271672954

The webpack rule looks like this -
```js
  {
    test: /\.(css)$/,
    use: [{
      loader: 'style-loader',
    }, {
      loader: 'css-loader',
      options: {
        sourceMap: true // toggle this flag to see the issue
      }
    }]
  },
```

I've bangged my head around this one for a long time and couldn't come up with anything. I'd be very glad if someone could point me in the right direction.

Webpack: 2.3.3
css-loader: 0.27.3

**UPDATE**:
Not sure if it helps, but notice this: 
`stroke='rgba(0, 0, 0, 0.5)'` // before
`stroke='rgba(0, 0, 0, 0.5")'` // after
This extra quote sign can't be good...

evilebottnawi:
[...]

PR #491, opened 2017-04-17
test: add tests when css contain data uri and source map enabled

evilebottnawi:
Tests

yes

not required

**Summary**

Ref: https://github.com/webpack-contrib/css-loader/issues/484

No

**Other information**

not required

### DIFF
--- a/test/sourceMapTest.js
+++ b/test/sourceMapTest.js
@@ -44,6 +44,40 @@ describe("source maps", function() {
 			version: 3
 		}]
 	]);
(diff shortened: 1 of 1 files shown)


IdanCo:
Wow, kudos for the quick reply and the excellent test!

I'm reopening this because for some reason (not sure how i missed it before) this only reproduces when sass-loader is present. Now, i now the reasonable thing would be to head out to their repo and report this issue, but before i do so there is still something strange the seems to originate from css-loader: **the sourcemap flag** still switches this bug on and off.

I've created a demo branch - https://github.com/IdanCo/webpack-modular/tree/inline-images
notice [this line here](https://github.com/IdanCo/webpack-modular/blob/inline-images/webpack.config.js#L27), it's what switches the bug on and off.

Any chance you can take a look at it? even if it's a sass-loader issue, the reason css-loader manifests it may be relevant. If there's any further information i can supply feel free.

Issue #459, opened 2017-03-17
Escape Characters in CSS content not processed correctly


rjnair-nm:
I am using css-loader and i could see that the content is not being processed correctly.

Original CSS :- 
```
.icon-caret-left:before {
   content: “\F10C"
}
```

The above CSS was processed by webpack to :- 
```
.icon-caret-left:before {
   content: '"\\f10c"'
}
```

Webpack: 2.2.1
CSS-Loader: 0.27.3
Node: 7.2.1

michael-ciniawsky:
@rjnair-nm for getting started did an earlier version of `css-loader` process it correctly? 😛

rjnair-nm:
@michael-ciniawsky  I had tried using css-loader 0.25.0 with webpack 1 and that used to work fine. I was upgrading to webpack 2 and while using the latest version of css-loader with that, i came across this issue.

michael-ciniawsky:
@rjnair-nm While copy and paste, maybe typo while you opened the issue maybe the root cause 😆 

```css
.icon-caret-left:before {
   content: “\F10C"
}
```
```
content: /* “  <= */\F10C" => “ => "
```

michael-ciniawsky:
The string delimiters don't match

michael-ciniawsky:
Otherwise I need webpack 1 && webpack 2 configs + ideally a small test repo for each to debug 😛

rjnair-nm:
@michael-ciniawsky I am using a custom library like Bootsrap which has the class definition as below :- 
```
.icon-caret-left:before {
  content: "\f10c"; 
}
```

When the webpack processes this via css-loader, it transforms to : -

```
.icon-caret-left:before {
   content: '"\\f10c"'
}
```

Webpack rule for css :-
```
const styleLoaders = ['css-loader?minimize&sourceMap&importLoaders=5', 'postcss-loader', 'sass-loader?sourceMap'];

{
              test: /.s?css$/,
              include: [path.join(root, 'src/client'), path.resolve(root, 'node_modules')],
              loader: IS_DEV ? 'style-loader!' + styleLoaders.join('!') : ExtractTextPlugin.extract({ fallback: 'style-loader', use: styleLoaders })
 }
```

I see this issue https://github.com/webpack-contrib/css-loader/issues/87 which was raised here. But in that case it was removing the escape characters, but in my case its adding extra escape characters :smile:

michael-ciniawsky:
[...]

PR #493, opened 2017-04-17
test: add test with escaped characters

evilebottnawi:
Tests

it is tests

not required

**Summary**

Ref: https://github.com/webpack-contrib/css-loader/issues/459

no

**Other information**

not required

### DIFF
--- a/test/simpleTest.js
+++ b/test/simpleTest.js
@@ -18,18 +18,29 @@ describe("simple", function() {
 	test("simple2", ".class { a: b c d; }\n.two {}", [
 		[1, ".class { a: b c d; }\n.two {}", ""]
 	]);
+	test("escape characters (uppercase)", ".class { content: \"\\F10C\" }", [
+		[1, ".class { content: \"\\F10C\" }", ""]
+	]);
+	// Need uncomment after resolve https://github.com/css-modules/postcss-modules-local-by-default/issues/108
+	/*test("escape characters (lowercase)", ".class { content: \"\\f10C\" }", [
+		[1, ".class { content: \"\\f10C\" }", ""]
+	]);*/
+	// Need uncomment after resolve https://github.com/mathiasbynens/cssesc/issues/10
+	/*test("escape characters (two)", ".class { content: \"\\F10C \\F10D\" }", [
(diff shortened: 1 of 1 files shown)


test/simpleTest.js
```diff
@@ -18,18 +18,29 @@ describe("simple", function() {
 	test("simple2", ".class { a: b c d; }\n.two {}", [
 		[1, ".class { a: b c d; }\n.two {}", ""]
 	]);
+	test("escape characters (uppercase)", ".class { content: \"\\F10C\" }", [
+		[1, ".class { content: \"\\F10C\" }", ""]
+	]);
+	// Need uncomment after resolve https://github.com/css-modules/postcss-modules-local-by-default/issues/108
```
michael-ciniawsky: Not sure about these tbh, maybe better keep a note and add them when it is resolved :)

evilebottnawi: @michael-ciniawsky i think best note is not in code, some issue can resolve in a long time :sob: Be good merge this as is and if some submit about problem related to this add link to issue in other repo maybe it is help solve the problem faster

michael-ciniawsky:
I will update the test suite today, so we can test within a 'real world' scenario (`ETWP` && `style-loader`)

evilebottnawi:
@michael-ciniawsky If I'm not mistaken in `postcss` `v6` we have builin `value-parse` maybe we can migrate to this, `postcss` more stable than other tools for this :smile: Now i think we can leave it at that.

michael-ciniawsky:
Instead of `css-selector-tokenizer`, I wanted to try taht ou but hadn't the time to do so until now, if `postcss-value-parser || postcss v6` could replace it I'm all in

evilebottnawi:
@michael-ciniawsky yep, I generally see for the first time the project using this module, we tried it in other projects, it did not meet expectations :sob: let's wait `postcss` `v6`, btw we can migrate on `postcss-value-parser` (or `postcss-values-parser`) it is more stable

michael-ciniawsky:
👍  I will open an issue for general overview with all available options here soon, so we can then triage this, but very good catch I'm surprised to be honest 😛

PR #496, opened 2017-04-17
feat: add externals support for CSS imports

fahad19:
## What kind of change does this PR introduce?

New feature via `externals` option in `css-loader`, much like Webpack's core externals feature.

## Did you add tests for your changes?

Yes.

## If relevant, did you update the README?

Yes.

## Summary

I opened an issue in `css-modules` repository for this first: https://github.com/css-modules/css-modules/issues/227

To summarise, I want a setup like this with [FrintJS](https://frint.js.org):

```html
<!-- vendors -->
<script src="lodash.js"></script>
<script src="react.js"></script>
<script src="frint.js"></script>

<!-- css-modules based framework, exposed as `window.CssFramework` -->
<script src="css-framework-using-css-modules.js"></script>

<!-- these Frint apps re-using vendors and CSS Framework from above -->
<script src="root-frint-app.js"></script>
<script src="child-frint-app-1.js"></script>
<script src="child-frint-app-2.js"></script>
```

Basically, I don't want Frint App bundles to include any CSS Framework code, since it has already been loaded before.

And my Webpack config for bundling Frint Apps would look like this for `css-loader`:

```js
{
  test: /\.css$/,
  exclude: /node_modules/,
  use: [
    'style-loader',
    {
      loader: 'css-loader',
      options: {
        modules: true,
        importLoaders: 1,
        localIdentName: '[name]__[local]___[hash:base64:5]',

        // this is the new feature
        externals: {
          'css-framework/variables.css': 'CssFramework.variables'
        },
      }
    },
  ]
}
```

In any of the App bundles, when I do this:

```js
import styles from './styles.css';
```

And my `styles.css` imports something from `css-framework` package:

```css
@value variables: "css-framework/variables.css";
@value primary-color, secondary-color from variables;

.button {
  color: primary-color;
}
```

Instead of including the `css-framework` code in the bundle, it would get the values dynamically in the browser from the global `window.CssFramework.variables`.

Possibly for old node versions, but I am happy to update them.

## Other information

Fully working proof-of-concept with [FrintJS](https://frint.js.org) here: https://github.com/fahad19/frint-css-poc

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -4,10 +4,23 @@
 */
 var loaderUtils = require("loader-utils");
 var processCss = require("./processCss");
+var getExternals = require("./getExternals");
 var getImportPrefix = require("./getImportPrefix");
 var compileExports = require("./compile-exports");
 var createResolver = require("./createResolver");
 
+function findExternalFromImportUrl(importUrl, externals) {
+	var externalsKeys = Object.keys(externals);
+	var key = externalsKeys.find(function (externalsKey) {
+		return importUrl.endsWith(externalsKey);
+	});
+
+	if (key) {
+		return externals[key];
+	}
+
+	return false;
+}
 
 module.exports = function(content, map) {
 	if(this.cacheable) this.cacheable();
@@ -18,36 +31,7 @@ module.exports = function(content, map) {
 	var camelCaseKeys = query.camelCase || query.camelcase;
 	var sourceMap = query.sourceMap || false;
 	var resolve = createResolver(query.alias);
-
-	var givenExternals = query.externals || (this.externals || {});
-	var externals = {};
-	if (Array.isArray(givenExternals)) {
-		givenExternals.forEach(function (external) {
-			if (typeof external === 'object') {
-				externals = Object.assign({}, externals, external);
-			}
-		});
-	} else {
-		externals = givenExternals;
-	}
-
-	if (typeof externals === 'string') {
-		externals = JSON.parse(externals);
-	}
-
-	var externalsKeys = Object.keys(externals);
-
-	function findExternalFromImportUrl(importUrl) {
-		var key = externalsKeys.find(function (externalsKey) {
-			return importUrl.endsWith(externalsKey);
-		});
-
-		if (key) {
-			return externals[key];
-		}
-
-		return false;
-	}
+	var externals = getExternals(this, query);
 
 	if(sourceMap) {
 		if (map && typeof map !== "string") {
@@ -90,7 +74,7 @@ module.exports = function(content, map) {
 			} else {
 				var importUrl = importUrlPrefix + imp.url;
 
-				var external = findExternalFromImportUrl(importUrl);
+				var external = findExternalFromImportUrl(importUrl, externals);
 				if (external === false) {
 					return "exports.i(require(" + loaderUtils.stringifyRequest(this, importUrl) + "), " + JSON.stringify(imp.mediaQuery) + ");";
 				}
@@ -105,7 +89,7 @@ module.exports = function(content, map) {
 			var importItem = result.importItems[idx];
 			var importUrl = importUrlPrefix + importItem.url;
 
-			var external = findExternalFromImportUrl(importUrl);
+			var external = findExternalFromImportUrl(importUrl, externals);
 			if (external === false) {
 				return "\" + require(" + loaderUtils.stringifyRequest(this, importUrl) + ").locals" +
 				"[" + JSON.stringify(importItem.export) + "] + \"";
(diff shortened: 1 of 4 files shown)


fahad19:
@bebraw: awesome! I will add docs + tests very soon and update here again :D

michael-ciniawsky:
Does this only work with when `{ modules: true }` (CSSModules) are enabled ?

fahad19:
@michael-ciniawsky: I cannot confirm it yet without tests. but manually, I only tested it with the `css-modules` approach, with `modules: true`.

michael-ciniawsky:
Where do the externals come from, could you give an example `externals` file please?

michael-ciniawsky:
That's important, since we plan to clearly separate CSSModules and CSS loading in the future due to perf problems in the current loader, if it works with modules exclusively we need to refactor that to `options.modules. externals` so it only gets trigger when in module 'mode'

bebraw:
@michael-ciniawsky If this is indeed CSS Modules specific, probably the smartest move would be to wait till the split happens and then push this functionality to the CSS Modules specific loader.

fahad19:
Externals are global objects. For e.g., if `css-framework.js` is loaded:

```html
<script src="css-framework.js"></script>
```

It would expose itself as `window.CssFramework`. The file would be generated from this source:

```js
// css-framework/index.js - entry for bundle
import variables from './variables.css';
import typography from './typography.css';

export {
  variables,
  typography,
};
```

And the CSS files would be:

```css
// css-frameworok/variables.css
@value primary-color: DodgerBlue;
@value secondary-color: DeepSkyBlue;

// css-framework/typography.css
.bold {
  font-weight: bold;
}

.italic {
  font-style: italic;
}
```

The object in the browser would look like this:

```json
// console.log(window.CssFramework);

{
  "variables": {
    "primary-color": "DodgerBlue",
    "secondary-color": "DeepSkyBlue"
  },
  "typography": {
    "bold": "typography__bold___2AHac",
    "italic": "typography__italic___3uoRc"
  }
}
```

And the final output of `css-framework.js` bundle: https://gist.github.com/fahad19/abc46027a48c738e6ff7ca8af04a696c

When other bundles do `require('css-framework')`-equivalent (from .css files), instead of bundling the CSS content, it would just get the values from global object in runtime.

fahad19:
currently, my needs are `css-modules` specific. I am happy to wait until the split happens. but if you think this feature can apply to both scenarios, that's even better I guess.

michael-ciniawsky:
Ahh 😛..  kk maybe I misunderstand things here, I though this is used especially for composition of [css] modules with 'compatible' frameworks. The feature is nice but we need to triage the scope first, `css-loader` will not run for modules by default anymore and therefore no `import/exports` will be genrated by default aswell only if `modules: true || {...options}`

fahad19:
updated the comment a bit more with examples: https://github.com/webpack-contrib/css-loader/pull/496#issuecomment-294724637

michael-ciniawsky:
kk I come back here as soon as I have a final clue about the other matters 😛

fahad19:
Here a fully working proof-of-concept repo: https://github.com/fahad19/frint-css-poc

fahad19:
* Tests added
* README updated
* PR body updated to reflect the last few commits

The feature being added does not require `modules: true`, as shown in the tests.

fahad19:
conflicts taken care of. up for final review. 

/cc @michael-ciniawsky @bebraw

lib/loader.js
```diff
@@ -19,6 +19,36 @@ module.exports = function(content, map) {
 	var sourceMap = query.sourceMap || false;
 	var resolve = createResolver(query.alias);
 
+	var givenExternals = query.externals || (this.externals || {});
```
bebraw: This logic would work better in a function of its own.

lib/loader.js
```diff
@@ -19,6 +19,36 @@ module.exports = function(content, map) {
...
+		externals = givenExternals;
+	}
+
+	if (typeof externals === 'string') {
+		externals = JSON.parse(externals);
+	}
+
+	var externalsKeys = Object.keys(externals);
+
+	function findExternalFromImportUrl(importUrl) {
```
bebraw: I would push this function outside to root scope.

michael-ciniawsky:
@fahad19 Also hold on with this a bit longer I have the groundwork for the next major of `css-loader` (no CSS Modules) ready locally and after that, I would like to move on with this feature here 😛

fahad19:
@bebraw: feedback taken care of.

@michael-ciniawsky: sure. please let me know when you are ready.

michael-ciniawsky:
Yep same as in #421 for the meantime

fahad19:
@d3viant0ne: thanks for the explanation!

just to clarify again, this new `externals` feature is **not** dependent on CSS Modules. it is intended to work with base css-loader.

michael-ciniawsky:
Sry but I have to declined this PR for now, since the `css-loader` v1.0.0 will be a complete overhaul starting with a minimal design

PR #497, opened 2017-04-17
perf: use modules only if explicitly set

evilebottnawi:
bugfix 

Already have.

After accept.

**Summary**

Now `css-loader` it is not loader for `css`, many project don't use `css-modules` for their project, but seems (https://github.com/webpack-contrib/css-loader/blob/master/lib/processCss.js#L152) they are use **always**. We spend a lot of time on `postcss-modules-*`, even if I set `modules: false`. I think this is completely wrong and many issue about performance probably related to this. Right name for this loader is `css-modules-loader`.

`postcss-loader` in v2 (i hope we can release it faster) wants to abandon `css-loader`, allow use only `postcss-loader` handle `css` and etc stuff (url, imports and etc). In fact, this is the real `css-loader`, It is simple, does not contain any logic.

What we can do?
1. Merge this PR before release `css-loader` (`css-modules-loader`) v1 and `postcss-loader` v2 (fix performance for project don't use `css-modules`).
2. Move all logic not related to `css-modules` (urls, improts and etc stuff) to `postcss-loader`.
3. Deprecated `css-loader` in favor `css-modules-loader`.
4. ...
5. PROFIT (better performance to project don't using `css-modules`, clear logic for `postcss-loader` and `css-modules-loader`, separation of issues - related to `css-modules` in `css-modules-loader`, related to `css` in `postcss-loader`)

Ref about `postcss-loader` `v2`: https://github.com/postcss/postcss-loader/issues/177

Should not.

**Other information**

Not required.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -14,7 +14,6 @@ module.exports = function(content, map) {
 	var callback = this.async();
 	var query = loaderUtils.getOptions(this) || {};
 	var root = query.root;
-	var moduleMode = query.modules || query.module;
 	var camelCaseKeys = query.camelCase || query.camelcase;
 	var resolve = createResolver(query.alias);
 
@@ -23,7 +22,6 @@ module.exports = function(content, map) {
 	}
 
 	processCss(content, map, {
-		mode: moduleMode ? "local" : "global",
 		from: loaderUtils.getRemainingRequest(this),
 		to: loaderUtils.getCurrentRequest(this),
 		query: query,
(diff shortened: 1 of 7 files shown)


michael-ciniawsky:
@evilebottnawi I take a deeper look later, but this is also related to the removal of 

```css
:local .className local(.className) {...}
```

then, which is the main reason to always use CSS Modules atm, but it's a bad practice anyways 😛

michael-ciniawsky:
@evilebottnawi Agian this would be definitely a breaking change ? Should be because of `:local(.className) {}`  not working anymore.
In this case we need to close this PR and move on as discussed, since the next semver major of `css-loader` will definitely **not** contain CSS Module s anymore 😛

evilebottnawi:
@michael-ciniawsky agree, let's close this and do all in steps :smile:

Issue #454, opened 2017-03-14
Unnecessary Buffer polyfill when source maps are disabled


alexwalter88:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
After upgrading from css-loader `0.26.4` to `0.27.3` webpack is including a polyfill for Buffer in the bundle when source maps are not enabled. This is resulting in an increased bundle size, even though the polyfill is never actually used.

**If the current behavior is a bug, please provide the steps to reproduce.**
Install css-loader `0.23.7`, run a webpack build, and check the resulting bundle.

**What is the expected behavior?**
If source maps are not enabled, then there is no need to output the `toComment` function in `lib/css-base.js`, which uses Buffer. The loader should only output this function if source maps are enabled.

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**
Windows 8.1
Node 6.9.5
Webpack 2.2.1

My current workaround is just adding
```
node: {
  Buffer: false
}
```
to the webpack config, but I think it would be better if the loader could handle this.

michael-ciniawsky:
@alexwalter88 https://github.com/webpack-contrib/css-loader/blob/master/lib/loader.js#L113 Normally it should exaclty do that, but guessing the ` toComment` function needs to be moved into the `if (useSourceMap)` statement to avoid including the Buffer polyfill

michael-ciniawsky:
Mind testing quickly testing that out ? 😛

alexwalter88:
Nope that doesn't seem to work. I tried removing the function too and just inlining it the `if (useSourceMap)` statement, but the Buffer polyfill is still being included.

michael-ciniawsky:
Alright I need to ask Tobias what to do about it at best, I come back when I know more :). Meanwhile please continue using your workaround 😛

bebraw:
I guess the question is, how to rewrite

```
var base64 = new Buffer(JSON.stringify(sourceMap)).toString('base64');
```

without relying on `Buffer`. That would be one way.

bebraw:
See [MDN](https://developer.mozilla.org/en/docs/Web/API/WindowBase64/Base64_encoding_and_decoding). Either `btoa` or solution `#1` from that page would likely solve this.

michael-ciniawsky:
Closed by #501

PR #501, opened 2017-04-20
fix: use `btoa` instead `Buffer`

evilebottnawi:

bugfix

yes

not required

**Summary**

https://github.com/webpack-contrib/css-loader/issues/454

no

**Other information**

not required

### DIFF
--- a/lib/css-base.js
+++ b/lib/css-base.js
@@ -54,7 +54,7 @@ function cssWithMappingToString(item, useSourceMap) {
 		return content;
 	}
 
-	if (useSourceMap) {
+	if (useSourceMap && typeof btoa === 'function') {
 		var sourceMapping = toComment(cssMapping);
 		var sourceURLs = cssMapping.sources.map(function (source) {
 			return '/*# sourceURL=' + cssMapping.sourceRoot + source + ' */'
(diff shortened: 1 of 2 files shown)


michael-ciniawsky:
Fixes #454

evilebottnawi:
@michael-ciniawsky i think to version `1` we should close all `bug` issues :smile:

evilebottnawi:
@michael-ciniawsky and be good release one patch version before `v1` it is allow catch all regression errors, if of course they will be :smile:

test/cssBaseTest.js
```diff
@@ -46,4 +64,17 @@ describe("css-base", function() {
 		}]);
 		m.toString().should.be.eql("body { a: 1; }\n/*# sourceURL=webpack://./path/to/test.scss */\n/*# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJmaWxlIjoidGVzdC5zY3NzIiwic291cmNlcyI6WyIuL3BhdGgvdG8vdGVzdC5zY3NzIl0sIm1hcHBpbmdzIjoiQUFBQTsiLCJzb3VyY2VSb290Ijoid2VicGFjazovLyJ9 */");
 	});
+	it("should toString without source mapping if btoa not avalibale", function() {
```
webdiscus, line 67: Why CSS should be extracted without source mapping if btoa not avalibale?
This is BREAKING CHANGE. In node <= 14 the CSS  has not the inline source map.
If btoa not avalibale then shoul be used the `Buffer.from(string).toString('base64')` instead of `btoa(string)`.

lib/css-base.js
```diff
@@ -54,7 +54,7 @@ function cssWithMappingToString(item, useSourceMap) {
 		return content;
 	}
 
-	if (useSourceMap) {
+	if (useSourceMap && typeof btoa === 'function') {
```
webdiscus, line 57: BREAKING CHANGE for node <= 14!
Extracting of source map SHOULD NOT be ignored!
Use the `Buffer.from(string).toString('base64')`

alexander-akait: This code for browser

test/cssBaseTest.js
webdiscus: ok, I understand.

P.S: I use the css-loader runtime in a webpack plugin to generate inline source-map. In node 16 it works, but not in older node.

Issue #463, opened 2017-03-22
url('') is rewritten as require('./')


darrenscerri:
**What is the current behavior?**
`url('')` is rewritten as `require('./')`

**If the current behavior is a bug, please provide the steps to reproduce.**
Simply specify a `url('')` rule in CSS.

**What is the expected behavior?**
`url('')` should not be allowed. Show a warning similar to https://github.com/webpack/webpack/issues/2006

michael-ciniawsky:
Yep, also`@ìmport/url()` with protocol schema `https://`, but we need to clearify if that breaks current useage somehow (don't think so, but how knows for sure 😛)

evilebottnawi:
Done in https://github.com/webpack-contrib/css-loader/pull/513

PR #513, opened 2017-04-23
fix: don't handle empty @import and url()

evilebottnawi:

bugfix

yes

not required

**Summary**

Ref: https://github.com/webpack-contrib/css-loader/issues/463

**Other information**

not required

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -161,18 +161,18 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
 	var pipeline = postcss([
 		localByDefault({
 			mode: options.mode,
-			rewriteUrl: function(global, origUrl) {
+			rewriteUrl: function(global, url) {
 				if(parserOptions.url){
-					var url = origUrl.trim(" ");
+                    url = url.trim();
 
-					if(!origUrl.replace(/\s/g, '').length || !loaderUtils.isUrlRequest(url, root)) {
-						return origUrl;
+					if(!url.replace(/\s/g, '').length || !loaderUtils.isUrlRequest(url, root)) {
+						return url;
 					}
 					if(global) {
 						return loaderUtils.urlToRequest(url, root);
 					}
 				}
-				return origUrl;
+				return url;
 			}
 		}),
 		extractImports(),
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -49,6 +49,9 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 				} else if(url.type === "string") {
 					url = url.value;
 				} else throw rule.error("Unexpected format" + rule.params);
+				if (!url.replace(/\s/g, '').length) {
+					return;
+				}
 				values.nodes[0].nodes.shift();
 				var mediaQuery = Tokenizer.stringifyValues(values);
 				if(loaderUtils.isUrlRequest(url, options.root) && options.mode === "global") {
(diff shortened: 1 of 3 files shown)


evilebottnawi:
I think we don't throw error on empty `url` and `import` because we don't know why it is used. Example i can create class with empty `.myclass { background: url('') }` and then change value on `js`.

michael-ciniawsky:
I will take a look soon 😛

michael-ciniawsky:
Fixes #463

lib/processCss.js
```diff
@@ -158,18 +161,18 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
...
-			rewriteUrl: function(global, url) {
+			rewriteUrl: function(global, origUrl) {
 				if(parserOptions.url){
-					url = url.trim(" ");
+					var url = origUrl.trim(" ");
 
-					if(!loaderUtils.isUrlRequest(url, root)) {
-						return url;
+					if(!origUrl.replace(/\s/g, '').length || !loaderUtils.isUrlRequest(url, root)) {
+						return origUrl;
```
michael-ciniawsky: ` return url ` ? Otherwise the former fix (`trim(" ")`) would be reverted, if you return the untrimmed `origUrl`, maybe better just `return url.trim(" ")` and rm `origUrl` 😛

evilebottnawi: @michael-ciniawsky i think we don't change original value of `url` (example `background: url('     ')` many spaces :smile: ) should be `background: url('     ')` after processing, if we return trimmed, we get `background: url('')` output (as now). I would not call it a bug, but this is a non-standard behavior that in theory we should not do.

michael-ciniawsky: What is the difference between `url(' ') && `url('  ')` ?

evilebottnawi: @michael-ciniawsky no difference :smile: but in theory i can have js script which get value url and check on ` `, I do not know why you need to do this and for what, but I can do it :smile: Just think that it's worth avoiding implicit manipulation that in the future can lead to unexpected problems

evilebottnawi: /cc @michael-ciniawsky ping :smile:

michael-ciniawsky: @evilebottnawi I'm not 💯 on this yet, mainly because of the code noise, even if small 😛 is justifiable, compared to what we want to achieve here. `var origUrl` in particular, the `trim()` story, Can't we solve this otherwise ? Is the distinction between `url('') <=> url('  ')` really *really* worthwhile? Besides that not handling empty `@import/url()` of course which needs to land :D

evilebottnawi: @michael-ciniawsky about `url` i think best solution if `url` empty (also with spaces and newlines and etc) do nothing at all, it is good practice don't change css ast, if we do not need it, but here in fact we do not need it. The same applies to `import`. We can never know why the user did this, but if he did it maybe he really needed it, and by changing something we can break his logic. If there are a lot of issues with this, then we can just throw `warning`, why warning? because it is not fatal, all works as expected :smile:

michael-ciniawsky: Maybe better to `trim()` and have the cleanest code possible + comment  and wait if someone needs the  `\s`  preserved and fix it **when** we have a usecase. I personally doubt that someone will ever need that, but in 3+ months I look at the code and wonder what the fuss was all about 😛

evilebottnawi: @michael-ciniawsky ok let's trim :smile:

evilebottnawi:
@michael-ciniawsky done, just see which tests changed and what the user has and what will get at the output https://github.com/webpack-contrib/css-loader/pull/513/commits/639bef177332bb6076edfb0ec90709ed00daae80#diff-aba9b3cb66ffd4aba9e67a4ad4d38376

lib/processCss.js
```diff
@@ -160,9 +163,9 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
 			mode: options.mode,
 			rewriteUrl: function(global, url) {
 				if(parserOptions.url){
-					url = url.trim(" ");
+                    url = url.trim();
```
michael-ciniawsky: Indent off? 🙃 😛

michael-ciniawsky:
> just see which tests changed and what the user has and what will get at the output 639bef1#diff-aba9b3cb66ffd4aba9e67a4ad4d38376

Is there something you're worried about besides the `url('  ') => url('  ')` case? Otherwise LTGM

evilebottnawi:
@michael-ciniawsky nothing :smile:

PR #517, opened 2017-04-25
docs(README): update import section

kud:
Gosh, I spent so much time understanding why I was able to compose a class with an internal class but not an external class in my css moduled style files.

I wanted to disable "import" because I didn't want that people use "@import" in the css files and wanted them make it importing via javascript like this:

```javascript
import '~/styles/index.css'
import styles from './index.css'
```

*BUT* disabling "import" option makes unavailable the fact to import external class in css-modules `composes` feature.

😒😒😒😒

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ To disable `@import` resolving by `css-loader` set the option to `false`
 @import url('https://fonts.googleapis.com/css?family=Roboto');
 ```
 
-> :waning: Use with caution, since this disables resolving for **all** `@import`s
+> :waning: Use with caution, since this disables resolving for **all** `@import`s, including css modules `composes: xxx from 'path/to/file.css'` feature.
 
 ### [`modules`](https://github.com/css-modules/css-modules)
 


michael-ciniawsky:
@kud I need to investigate into this and why, tbh I also thought it just disables `@import` resolving 😒

kud:
For the moment, I was able to 

```css
.class {
  composes: otherClass
}
```

but not

```css
.class {
  composes: otherClass from 'other.css'
}
```

when `import: false`.


Oh and by the way, is it normal that you can do:


```css
.otherClass {}

.class {
  composes: otherClass
}
```

but not:

```css
.class {
  composes: otherClass
}

.otherClass {}
```

?

michael-ciniawsky:
CSS Modules use a format called ICSS (Intermediate CSS) `:import, :export`basically a JSON Blob

```js
{
   "className": "dfbzreur474"
   ...
}
```

For the first, I will check that I assumed that `composes` actually loads the file **independently**, modularizes it and you can compose it then (external, like you mentioned)

For the second, it is 'correct' in the current plugin order used by `css-loader` the `.otherClass {}` isn't available (parsed) when the `compose` looks up the ICSS representation


We will remove CSS Modules from `css-loader` with the next major release and release CSS Modules as 'normal' PostCSS Plugin usable with `postcss-loader` as opt-in and better configuration for the important parts of modules 😛 but it's WIP atm

kud:
Okay, thank you for your feedback. It makes sense not to have css modules in css-loader but in a postcss as plugin. Yes. :)

michael-ciniawsky:
I will merge this, when I have certainty of what is really going on :)

michael-ciniawsky:
Ooops... 😛

kud:
Thank you @michael-ciniawsky, It'll maybe help some people before you remove css modules from here or patch it.

Issue #317, opened 2016-08-12
utf-8 characters support?


reverland:
I try to use css-loader to load this file

``` css
body {
  font-family: '微软雅黑'; /* some chinese font name */
}
```

but get this

``` javasript
    exports.push([module.id, "@charset 'UTF-8';\nbody {\n  font-family: '\\5FAE\\8F6F\\96C5\\9ED1';\n}\n", ""]);
```

It looks like the font name is transformed into something else.

follow the way i found `stringfyValues` which will `stringify` the utf-8 character...

https://github.com/css-modules/css-selector-tokenizer/blob/master/lib/stringifyValues.js

what the declaration value is stringified for? Is there any way to use non ascii font name with css-loader?

imingyu:
font-family:Microsoft Yahei

japboy:
Please support non-ascii characters in css-loader 👍

swcho:
For me, this problem seems hard to solve at a first glance as unicode escaping done by postcss module.

Until proper solution applied, I would you interim solution by using hacky loader which unescape unicode.
https://github.com/swcho/font-family-unescape-loader

You can place "font-family-unescape-loader" right before "css-loader" then the loader will change font names of font-family config.

evilebottnawi:
WIP

evilebottnawi:
oh, two problems:
1. `postcss-modules-local-by-default` change `微软雅黑` to `\5FAE\8F6F\96C5\9ED1`.
2. `css-loader` on line https://github.com/webpack-contrib/css-loader/blob/master/lib/processCss.js#L129 also change `微软雅黑` to `\5FAE\8F6F\96C5\9ED1`.

First try to fix in `css-loader`, then in `postcss-modules-local-by-default`

evilebottnawi:
Close in favor https://github.com/webpack-contrib/css-loader/issues/578

PR #523, opened 2017-04-27
[WIP] fix: utf-8 characters support, emoji also 😸

evilebottnawi:
bugfix

yes

not required

**Summary**

Ref: https://github.com/webpack-contrib/css-loader/issues/317

Also we should fix incorrect behvaior with utf-8 characters in `postcss-modules-local-by-default`. Let's leave this PR here, after how i resolve problem in `postcss-modules-local-by-default` we can merge this issue.

no

**Other information**

### DIFF
--- a/test/importTest.js
+++ b/test/importTest.js
@@ -9,18 +9,30 @@ describe("import", function() {
 	], "", {
 		"./test.css": [[2, ".test{a: b}", ""]]
 	});
+	test("import 2", "@import url('test.css');\n.class { a: b c d; }", [
+		[2, ".test{a: b}", "screen"],
+		[1, ".class { a: b c d; }", ""]
+	], "", {
+		"./test.css": [[2, ".test{a: b}", "screen"]]
+	});
+	test("import non ascii characters (emoji)", "@import url(😸.css);\n.class { a: b c d; }", [
+		[2, ".test{a: b}", ""],
+		[1, ".class { a: b c d; }", ""]
+	], "", {
+		"./😸.css": [[2, ".test{a: b}", ""]]
+	});
 	test("import camelcase", "@IMPORT url(test.css);\n.class { a: b c d; }", [
 		[2, ".test{a: b}", ""],
 		[1, ".class { a: b c d; }", ""]
 	], "", {
 		"./test.css": [[2, ".test{a: b}", ""]]
 	});
-    test("import empty url", "@import url();\n.class { a: b c d; }", [
-        [1, "@import url();\n.class { a: b c d; }", ""]
-    ], "");
-    test("import empty url with quotes", "@import url('');\n.class { a: b c d; }", [
-        [1, "@import url('');\n.class { a: b c d; }", ""]
-    ], "");
+	test("import empty url", "@import url();\n.class { a: b c d; }", [
+		[1, "@import url();\n.class { a: b c d; }", ""]
+	], "");
+	test("import empty url with quotes", "@import url('');\n.class { a: b c d; }", [
+		[1, "@import url('');\n.class { a: b c d; }", ""]
+	], "");
 	test("import with string", "@import \"test.css\";\n.class { a: b c d; }", [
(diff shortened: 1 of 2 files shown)


evilebottnawi:
/cc @michael-ciniawsky 
why?
1. `css-selector-tokenizer` don't works correctly with string, why they always escape this?
2. `css-selector-tokenizer` don't supports non ASCII characters (national characters, emoji :sob: ).
2. `css-selector-tokenizer` uses regex, it is bad, we should use parser.
3. Using `postcss-value-parser` we take more flexible code.
4. `postcss-value-parser`` more faster then `css-selector-tokenizer` in most use case (just try to use `console.time` :smile: ).
5. ...
6. Profit

michael-ciniawsky:
👍 I'm all in removing it 

6. ...
7. Profit

😆 🤣 lol...

michael-ciniawsky:
> It makes sense when there is...

=> It makes **no** sense when there is... ?

Yep let's wait, what comes up there 🙃

michael-ciniawsky:
@evilebottnawi Still WIP 🙃 ?

evilebottnawi:
@michael-ciniawsky This PR is ready to merge, but need patch in `css-modules` plugins, let's leave it here for now, i plan to solve the problem in near future.

michael-ciniawsky:
Ahh right you mentioned that 😛  I forgot sry yep kk

michael-ciniawsky:
Can we close and fix it in `css-modules`and `develop`? Ping @ sullenor or @ TrySound which are maintainers in `css-modules` now. Also join https://gitter.im/css-modules/maintainers in case we need someone from `css-modules´ :D

evilebottnawi:
@michael-ciniawsky we should merge this and fix it in `css-modules` (if we fix it only in `css-modules` it this will leave the problem). Maybe join is good :smile:

evilebottnawi:
@TrySound i think you can help our with this :smile: Now we have `css-selector-tokenizer` in `css-loader` (also in `postcss-modules-local-by-default`, `postcss-modules-scope` and `postcss-modules-resolve-imports`). Should we rewrite everything on `postcss-value-parser` instead `css-selector-tokenizer`, or will correct errors in `css-selector-tokenizer` (now he doesn't support emoji and national language, also related to bug with `content: "\F10C \F10D"` https://github.com/mathiasbynens/cssesc/issues/10)? 

Bug here https://github.com/css-modules/css-selector-tokenizer/blob/master/lib/stringifyValues.js#L7 or maybe not bug, but it is not works as expected. You can find several issues with this behavior here.

TrySound:
@evilebottnawi Did you have failed tests?

TrySound:
I can't reproduce this problem because node.js can't parse this symbol
```
    fixture: '.foo { content: '😸 😸'; }'
                               ^
SyntaxError: Unexpected token ILLEGAL
```

evilebottnawi:
@TrySound just take tests from here and run on current master branch, maybe someone has already fixed this, but the error was

evilebottnawi:
@TrySound Original issue https://github.com/webpack-contrib/css-loader/issues/317

jo12bar:
Is this still WIP? I keep running into the issue described in #526 when trying to use faceyspacey/extract-css-chunks-webpack-plugin.

evilebottnawi:
@jo12bar yep, wip, in near future I'll finish it

michael-ciniawsky:
@evilebottnawi What's the status of this PR, should we land this before `v1.0.0` ?

evilebottnawi:
@michael-ciniawsky yes after 1.0.0

Pro-Novice:
What kind of proposal do I submit for GSoC? 

I have a few Node.js Applications .... Will that be of any help ?

michael-ciniawsky:
Status ? :)

evilebottnawi:
Close in favor https://github.com/webpack-contrib/css-loader/pull/748

Issue #529, opened 2017-05-09
source-map resource urls coming out strange


DBosley:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Using postcss-loader (only using autoprefixer plugin, no extra configurations) combined with sass-loader, css-loader, and style-loader causes the resource URL for sourcemaps to be strange:

example: http://imgur.com/a/R73hV
scss loader config
```
{
          test: /\.scss$/,
          use: [{
              loader: 'style-loader',
              options: {
                sourceMap: true,
              }
            },
            {
              loader: 'css-loader',
              options: {
                sourceMap: true,
              }
            },
            {
              loader: 'postcss-loader',
              options: {
                sourceMap: true,
                //parser: require('postcss-scss'),
              }
            },
            {
              loader: 'sass-loader',
              options: {
                includePaths: [
                  './src/styles/bootstrap',
                  './src/styles/partials'
                ],
                sourceMap: true,
              }
            },
          ],
          include: [helpers.root('src', 'styles')]
        }
```

**What is the expected behavior?**
relative urls for sourcemaps, or at least only one absolute url

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**
webpack: 2.5.1
node: 6.10.3
windows 10

DBosley:
This issue seems to affect multiple loaders. Posted the same issue on all affected loaders:
https://github.com/webpack-contrib/css-loader/issues/529
https://github.com/webpack-contrib/sass-loader/issues/451
https://github.com/webpack-contrib/style-loader/issues/231
https://github.com/postcss/postcss-loader/issues/224

michael-ciniawsky:
It's likely to be in `postcss-loader` :)

DBosley:
I tried quite a few different configs today trying to figure this out. Some for scss files, some for css. For css files, I wasn't ever able to get relative paths in chrome, but by ordering my css config like this:
```
{
  test: /\.css$/,
  use: [{
      loader: 'style-loader',
    },
    {
      loader: 'css-loader',
      options: {
        sourceMap: true
      }
    },
    {
      loader: 'postcss-loader',
    }
  ],
  include: [helpers.root('src', 'styles')]
},
```
I was able to get rid of the extra absolute paths.

michael-ciniawsky:
Should be fixed by #532, released in `v0.28.2`

PR #532, opened 2017-05-11
fix: source maps path on `windows`

evilebottnawi:
bugfix

yes

**Summary**

https://github.com/webpack-contrib/css-loader/issues/529

No.

**Other information**

We need more tests and good review on this.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -20,8 +20,17 @@ module.exports = function(content, map) {
 	var resolve = createResolver(query.alias);
 
 	if(sourceMap) {
-		if (map && typeof map !== "string") {
-			map = JSON.stringify(map);
+		if (map) {
+			if (typeof map !== "string") {
+				map = JSON.stringify(map);
+			}
+
+			if (map.sources) {
+				map.sources = map.sources.map(function (source) {
+					return source.replace(/\\/g, '/');
+				});
+				map.sourceRoot = '';
+			}
 		}
 	} else {
 		// Some loaders (example `"postcss-loader": "1.x.x"`) always generates source map, we should remove it
@@ -30,8 +39,8 @@ module.exports = function(content, map) {
 
 	processCss(content, map, {
 		mode: moduleMode ? "local" : "global",
-		from: loaderUtils.getRemainingRequest(this),
-		to: loaderUtils.getCurrentRequest(this),
+		from: loaderUtils.getRemainingRequest(this).split("!").pop(),
+		to: loaderUtils.getCurrentRequest(this).split("!").pop(),
 		query: query,
 		minimize: this.minimize,
 		loaderContext: this,
@@ -105,11 +114,11 @@ module.exports = function(content, map) {
 			map = result.map;
 			if(map.sources) {
 				map.sources = map.sources.map(function(source) {
-					return source.split("!").pop();
+					return source.split("!").pop().replace(/\\/g, '/');
 				}, this);
 				map.sourceRoot = "";
 			}
-			map.file = map.file.split("!").pop();
+			map.file = map.file.split("!").pop().replace(/\\/g, '/');
 			map = JSON.stringify(map);
 			moduleJs = "exports.push([module.id, " + cssAsString + ", \"\", " + map + "]);";
 		} else {
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -21,7 +21,7 @@ module.exports = function(content, map) {
 
 	if(sourceMap) {
 		if (map) {
-			if (typeof map !== "string") {
+			if (typeof map === "string") {
 				map = JSON.stringify(map);
 			}
 
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -48,7 +48,7 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 					url = url.url;
 				} else if(url.type === "string") {
 					url = url.value;
-				} else throw rule.error("Unexpected format" + rule.params);
+				} else throw rule.error("Unexpected format " + rule.params);
 				if (!url.replace(/\s/g, '').length) {
 					return;
 				}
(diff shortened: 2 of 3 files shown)


lib/loader.js
```diff
@@ -22,6 +22,14 @@ module.exports = function(content, map) {
 	if(sourceMap) {
 		if (map && typeof map !== "string") {
 			map = JSON.stringify(map);
+
+			if (map.sourceRoot) {
+				map.sourceRoot = '';
+			}
+
+			if (map.sources) {
+				map.sources = map.sources.map((source) => source.replace(/\\/g, '/'));
```
michael-ciniawsky: `.map((source) => ...)` => `.map(function(source) { return ... })`

michael-ciniawsky:
https://github.com/postcss/postcss/blob/master/lib/map-generator.es6#L188-L189

But it seems to work on the relative file path (L187) only atm and maybe unrelated 😛

lib/loader.js
evilebottnawi: @michael-ciniawsky oh yep, i don't support old node.js in my project :smile:

michael-ciniawsky: Neither do I 😛 and I'm really favour make the switch as quickly as possible (node `< 4.0.0` is dead), but if we 'break' things all kind of conservatives will come out of their rat holes  and complain like crazy about it :D 'breaks app IN PRODUCTION' 'USE SEMVER!!!` etc...

lib/loader.js
```diff
@@ -30,8 +39,8 @@ module.exports = function(content, map) {
 
 	processCss(content, map, {
 		mode: moduleMode ? "local" : "global",
-		from: loaderUtils.getRemainingRequest(this),
-		to: loaderUtils.getCurrentRequest(this),
+		from: loaderUtils.getRemainingRequest(this).split("!").pop(),
```
michael-ciniawsky: What is the output of `loaderUtils.getRemainingRequest(this).split("!").pop()`? If I remember right it should be the absolute file path and therefore we could replaced it with `this.resourcePath` instead?

lib/loader.js
```diff
@@ -30,8 +39,8 @@ module.exports = function(content, map) {
...
+		to: loaderUtils.getCurrentRequest(this).split("!").pop(),
```
michael-ciniawsky: Should be equal to `this.resourcePath` aswell

lib/loader.js
evilebottnawi: @michael-ciniawsky we can have `remainingRequest` https://github.com/webpack/loader-utils/blob/master/lib/getRemainingRequest.js#L4

evilebottnawi: Best get file from `getRemainingRequest` it is more safe

michael-ciniawsky: kk also not really related to this PR, just pop into my eyes

evilebottnawi:
@michael-ciniawsky if we enable `appveyor` i can rewrite all tests for compatibility with `windows` (now tests written only for linux based os)

lib/loader.js
```diff
@@ -20,8 +20,17 @@ module.exports = function(content, map) {
...
-		if (map && typeof map !== "string") {
-			map = JSON.stringify(map);
+		if (map) {
+			if (typeof map === "string") {
+				map = JSON.stringify(map);
+			}
+
+			if (map.sources) {
+				map.sources = map.sources.map(function (source) {
+					return source.replace(/\\/g, '/');
```
michael-ciniawsky: Test with `path.sep`, but I think it's unnecessary...

michael-ciniawsky:
Yep appveyor would be good idea, want is need to enable it besides `appveyor.yml` ? For Org access ping bebraw or d3viant0ne 😛

lib/loader.js
evilebottnawi: @michael-ciniawsky yep, we should already changed `\` to `/`

evilebottnawi:
@michael-ciniawsky i think we should do this in `webpack-default` (no time for this :sob: )

evilebottnawi:
@bebraw this https://github.com/webpack-contrib/css-loader/pull/532#discussion_r116264754 or https://github.com/webpack-contrib/css-loader/pull/532#discussion_r116258997 ?

lib/loader.js
bebraw: @michael-ciniawsky This is a source map check. Can we assume forward slash there due to browser environment?

lib/loader.js
bebraw: Look into `resourcePath`, yup. I can't remember the exact API, though. Maybe this needs some guidelines.

evilebottnawi: @bebraw why? what is bad use `loaderUtils.getRemainingRequest(this).split("!").pop()`, also https://github.com/webpack-contrib/css-loader/pull/532#discussion_r116260918

d3viant0ne:
@bebraw - Need to check with Tobias direction before merging anything into css-loader, he specifically requested the code churn be kept to a minimum while he is working on new-loader.

bebraw:
@d3viant0ne Yeah, fair point. Better get Tobias' approval first. 👍

lib/loader.js
bebraw: @evilebottnawi I'm not entirely sure. At least test-wise the current implementation passes. It's fine to refactor later. 👍

michael-ciniawsky:
@evilebottnawi @d3viant0ne @bebraw I think we can merge this meanwhile, despite the fact that current implementation will hopefully vanish soon 😛 , it's definitely a bug in the current `css-loader`

d3viant0ne:
@evilebottnawi Open an issue in defaults for appveyor and we can get started on a org wide config as a part of defaults.

d3viant0ne:
As far as this particular PR goes, there are a few open review comments. Patches are fine for the current major version so @bebraw @evilebottnawi @michael-ciniawsky if everyone is satisfied with this, throw a thumbs up on the comment and we'll get this landed.

michael-ciniawsky:
AppVeyor Tracking Issue [#54](https://github.com/webpack-contrib/webpack-defaults/issues/54)

PR #542, opened 2017-05-24
[develop] refactor: apply webpack-defaults

michael-ciniawsky:
## Refactor

- [x] apply `webpack-defaults`

## Features

- [x] pass PostCSS AST directly to avoid 2x parsing (perf)
- [x] pass `locals`

## Breaking

- [x] removed `options.root`
- [x] removed `options.minimize` (CSSNano)
- [x] removed `options.modules`
- [x] removed `options.camelCase`
- [x] removed `options.importLoaders`

## TODO

- [ ] allow to filter certain `url()`'s (`options.url` `{Array|Regex}`)
- [ ] readd CSS Modules
- [ ] readd `loclas` processing (`camelCase` as default, no option ?)
- [ ] export CSS && `locals` as ES2015 Module

## Discuss

- No `:locla :local()` (Global Mode) ? (Recommended)
- Readd `cssnano` or move to `postcss-loader` ? 

cc @sullenor @TrySound


src/index.js
```diff
@@ -0,0 +1,156 @@
...
+        map = JSON.stringify(map);
+
+        modules = `exports.push([module.id, ${css}, '', ${map}]);`;
+      } else {
+        modules = `exports.push([module.id, ${css}, '']);`;
+      }
+
+      let locals;
+
+      if (meta && meta.locals) {
```
michael-ciniawsky: This would be where locals are added if `postcss-loader` passes them, needs to be replaced with the current implementation

src/index.js
```diff
@@ -0,0 +1,156 @@
...
+    from: file,
+    to: file,
+    url: options.url,
+    import: options.import,
+    sourceMap: options.sourceMap
+  };
+
+  Promise.resolve()
+    .then(() => {
+      if (meta && meta.ast) return processor(ast, map, processOptions);
```
michael-ciniawsky: We can pass the PostCSS AST directly now to avoid a second parse of CSS

evilebottnawi: @michael-ciniawsky it be good, very good, more perf

src/index.js
evilebottnawi: @michael-ciniawsky I would prefer
> import css, { className1, className2 } from './file.css' {String}, {Locals}

But honestly I do not even know :smile:

evilebottnawi:
@michael-ciniawsky But what about the new loader? Seems this PR related to current implementation.

Because we have problems with perf and css-modules. If i set `modules: false`, `postcss` plugins for `css-modules` still working (https://github.com/webpack-contrib/css-loader/blob/master/lib/processCss.js#L162), it is failed in perf. We should fix this in this PR.

test/configs/alias.json
```diff
@@ -0,0 +1,8 @@
+{
+  "file": "alias",
```
evilebottnawi: Please don't use config as json, this does not make it possible to test the js functions and other expressions

evilebottnawi:
Don't remove tests, we must adapt them, otherwise we will get millions of regressions as `postcss-loader`

test/configs/alias.json
evilebottnawi: And frankly speaking, this approach to testing is not very convenient, we need to revise it.

michael-ciniawsky: Yep, this needs an update as in `postcss-loader`

michael-ciniawsky:
@evilebottnawi I adpated a basic tests and a few are missing as the decision to move directly to `new-loader` was made I stopped working on this. If we start afresh with CSS Modules and PostCSS I will continue to port all tests and anyone can make PR against `develop`. This is just the baseline atm

michael-ciniawsky:
`modules` are removed atm, but @sullenor what's to readd them with a different implementation, we definitely need to avoid `:local :local()` in 'global' mode and test perf in general with modules

michael-ciniawsky:
The `develop` is a `PoC`, but we need a place to start and triage, either turns out bad or we can adapt what's in `new-loader` etc. and move on here

evilebottnawi:
@michael-ciniawsky seems we need good roadmap for `css-loader`, also decide
 question about `@import` and `url()` in `new-css-loader`.

michael-ciniawsky:
Yep I take a look at this PR later and I think we should just start then, all outstanding things can get there own issue and PR for discussion and we have at least `webpack-defaults` and cleaner code base, everything else follows... 😛

src/index.js
michael-ciniawsky: 😅 yeah.. it's a possiblity `import * as styles from 'file.css'` is the clear favourite 🏅

d3viant0ne:
Given defaults already exists in `new-loader` I don't know that it really makes sense to apply defaults to the existing css-loader. It really depends on what the plan is for the style related loader chain as a whole which hasn't been disseminated yet.

We have been asked to not do anything major with css-loader while Tobias is working on the new-loader and this definitely qualifies, ping him in slack and get his opinion on defaults as it relates to the existing css-loader.

src/index.js
```diff
@@ -0,0 +1,156 @@
...
+
+        modules = `exports.push([module.id, ${css}, '', ${map}]);`;
+      } else {
+        modules = `exports.push([module.id, ${css}, '']);`;
+      }
+
+      let locals;
+
+      if (meta && meta.locals) {
+        locals = `exports.locals = ${meta.locals};`;
```
sullenor: Should `meta.locals` be a string? 
I actually thought that user would pass a plain object here, so it will be convenient to use it :)

sullenor: Also, in the future versions the object may contain a prototype chain (https://github.com/css-modules/css-modules/issues/147), so it would be necessary to make a proper serializer here.

src/index.js
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @method loader
+ *
+ * @param  {String} src  Source
+ * @param  {Object} map  Sourcemap
+ * @param  {Object} meta Metadata
+ *
+ * @return {cb} cb       Callback
+ */
+export default function loader (src, map, meta) {
```
michael-ciniawsky: Yep this should work according to sokra, needs to be tested though :), the third argument of the callback can take a meta `{Object}`

```js
function loader (src, map, meta) {// {String} | {Object} | {Object}
   ....
   .then(({ css, map, messages }) => cb(null, css, map, { messages })
}

File: src/index.js (hunk 4 above)
Comment (michael-ciniawsky) [inline src/index.js]: This is more of a placeholder yet :D :D I was in the process of trying out how to ES2015 modulify this. That's where we need to start afresh and if CSSM tsay in `css-loader` there will be no meta locals, since this was originally intended to coma from `postcss-loader`. Any other ideas how we can do it a highly welcome 😛 

```
let locals = []

Object.keys(meta.locals).forEach((key) => {
     locals.push `export const ${key} = ${meta.locals[key]}`
})

locals = locals.join('\n')
```

michael-ciniawsky:
@d3viant0ne The `develop` branch is only a thing if CSSM stay in one way or the other, but since we don't really know how to handle CSS Modules (removal) atm and all the this and that won't move any further without some hands on code, we would like to try out the possibilities :). If we have certainty about CSSM we either ask Tobias the fuss here. `develop` can likely be 🔪  at the end of the day :D.

michael-ciniawsky:
@sullenor @evilebottnawi Should we just merge this into `develop` so everyone can start to make PR's against develop ? There's plenty of work left 😛

src/index.js
sullenor: `JSON.stringify` will be enough for the beginning :)
Actually, `css-loader` had an option, which camelized key names of the locals, so thats why it was a bit complicated and looks like some ppl still use it.

sullenor:
I think the right solution will be make a single postcss plugin, which will contain the necessary set of plugins which implement the `css-modules` spec and expose the necessary api. It will allow to embed (or to remove) it easily. Since our plan is not to make a major changes (as I understood @d3viant0ne right) this solution should fit here.

But there are some difficulties related to the css modules `composition` part. Composition (icss `:import()`) provides a possibility to load additional files. This part I would like to have also as a part of css-modules implementation (postcss plugin). However `css-loader` is almost a latest loader in the chain and for the proper behavior we need also to pass the loaded chunks through the remanning loaders in the chain to preprocess them.

sullenor:
> we need also to pass the loaded chunks through the remanning loaders in the chain to preprocess them

I guess `importLoaders` flag was designed for that purpose.

michael-ciniawsky:
Yeah agree it would be the best to have a standalone plugin `postcss-modules` overhaul 🙃 ? + e.g `result.messages['CSSM API here'] 

`importLoaders` prefix `@import` with a query so it becomes a request on it's own => `@import 'path/to/file.css'` =>  importLoaders => `require(-->./~css-loader?{}!./~postcss-loader?{}!<--path/to/file.css)` the idea is good the implementaion is bad :blush:

d3viant0ne:
@michael-ciniawsky - Tobias has a design for the style loader chain which he will be disseminating to all of us soon.

In the short term, the `right solution` is to do nothing until we know how css support in webpack is going to work, when it's going to land and how that effects the existing published loaders.

Issue #131, opened 2015-08-27
Import error with modules mode, composes and sass


cliffrowley:
I have a project in which I have a `Card` and `CardContainer` component.  The idea is that the `Card` can be used in isolation if desired, or multiple cards can be used inside a `CardContainer` which will arrange its cards in a grid and provide scrolling if necessary.  In order to facilitate this, the container's CSS needs to modify the flex attributes of its children (adding `flex 1 1 auto` for example).

I'm using sass (via sass-loader) and each of my component sass files imports a `utils.scss` a couple of directories down, which works fine without `composes`, but with it I am getting a file not found on the `utils.scss` from the file being imported for composition.

This is my directory structure:

```
src/
  components/
    Card.jsx
    Card.scss
    CardContainer.jsx
    CardContainer.scss
    ...
  sass/
    _utils.scss
    ...
```

My (simplified) `Card.scss` looks like this:

```
@import '../sass/utils';

.card {
  ...
}
```

My (again simplified) `CardContainer.scss` looks like this:

```
@import '../sass/utils';

.card {
    composes: card from './Card.scss';
}

.cardContainer {
  & > .card {
    ...
    margin: 0 $gutter $gutter 0;
    flex: 1 1 auto;
    width: 30%;
  }
}
```

The error I see is:

```
Module not found: Error: Cannot resolve 'file' or 'directory' ../sass/utils in <path to project>/src/components
```

I don't fully understand how this all fits together but It seems to me that this mechanism is trying to resolve a js/jsx file rather than going through the sass pipeline.  Does this make sense to you?  It's resolving `./Card.scss` just fine - it's the secondary import that doesn't seem to be working.

l-urence:
Have similar issue with stylus. It seems to be that the stylus loader from the configuration file is not getting called. In my case transforming the styl file within the compose helped.

```stylus
composes: formInput from 'stylus!./elements.styl'
```

In your case you have to call the sass loader instead of stylus.

cliffrowley:
Oh, that's a great idea for a workaround - I hadn't even considered that - thanks! Obviously this is still a bug (or lack of feature?) somewhere along the line but this will certainly help move things along for me and my project very nicely!

cliffrowley:
@l-urence your workaround worked for sass too, fyi - thank you so much :+1:

bkonkle:
[...]

PR #544, opened 2017-05-29
feat: add `options.matchImportLoadersByOrigin`

delijah:
This PR introduces a new config option called "matchImportLoadersByOrigin" config option. With this config option set to true, all imports (@import and composes) are resolved matching webpack loaders by it's origin instead of just using the same loader chain, that was used for it's containing file. importLoaders config option is not needed anymore, when using matchImportLoadersByOrigin.

https://github.com/webpack-contrib/css-loader/pull/287 is closely related and probably even the more solid version, but without config option. I do not understand why this never has been merged since it would solve so many issues.

If it is needed, i can add tests. First of all i would like to see, if my changes make sense.

Yes.

**Summary**
This should resolve the following issues:
https://github.com/webpack-contrib/css-loader/issues/286
https://github.com/webpack-contrib/css-loader/issues/131
https://github.com/css-modules/css-modules/issues/110
https://github.com/css-modules/css-modules/issues/170

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ It's useful when you, for instance, need to post process the CSS as a string.
 |**`sourceMap`**|`{Boolean}`|`false`|Enable/Disable Sourcemaps|
 |**`camelCase`**|`{Boolean\|String}`|`false`|Export Classnames in CamelCase|
 |**`importLoaders`**|`{Number}`|`0`|Number of loaders applied before CSS loader|
+|**`matchImportLoadersByOrigin`**|`{Boolean}`|`false`|Match import loaders by origin through webpack.config|
 
 ### `root`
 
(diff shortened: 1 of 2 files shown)


michael-ciniawsky:
@delijah Could we add a prop to `importLoaders` instead, or default to this behaviour if `importLoaders` isn't set by the user ? The changes don't really justify to introduce a new option :D 😛 . Also [this comment](https://github.com/webpack-contrib/css-loader/pull/287#issuecomment-261269199) bugs me, we need to clearify what he meant by that :)

sullenor:
@michael-ciniawsky should it be handled by `resolve.extentions`?
https://webpack.js.org/configuration/resolve/#resolve-extensions

sullenor:
ah, sorry, read it wrong. Actually `importLoaders` with proper number should solve it (you should set it to the number of loaders, that are after `css-loader` in the chain).

delijah:
@sullenor i think a lot of things are working somehow. But using the same loader (let's assume a loader rule created for handling css) for a completely different file type (let's assume sass) just feels wrong.

I just tried to reconstruct those errors, but i was not able to. Seems like there was something changed since those issues were created.

So no real problems, but i think this importLoaders config option still does make no sense. Why not just using standard webpack behaviour and use the defined loader rules from webpack.config?

What do you think?

michael-ciniawsky:
**entry.css**
```css
@import './style.css';

.className { color: red;  }
```

```
style-loader <= css-loader <= postcss-loader 
```

**after `css-loader`**
```
/* importPrefix === 'css!postcss!'
@import './style.css'  => require('${importPrefix}!./style.css') 
```

Without the `importPrefix`  `(`importLoaders`) the `@import` wont't be transpiled to my knowlegde, that's why it was added. The possibility to make a new request and it respects `module.rules` for that request would make `importLoaders` obsolete (witch would be a good thing) I'm not :100: on this either 

```
const css = [
  [module.id, css, map, ...] // @import
  [module.id, css, map, ...] // entry.css
]
```

Your change is to omit the prefix, but isn't that the default already ? I'm missing something here ? Please elaborate further on this :)

michael-ciniawsky:
@delijah Would this work, if `importLoaders` is set for other `@imports` ? I don't think this solves the issue with wrong request in `composes` in case it's composed from a different file format and other `@import`'s relying on `importLoaders` wouldn't work anymore? It would require additional logic to omit the `importPrefix` for files where it isn't needed

PR #559, opened 2017-06-21
docs(README): `importLoaders` should be 2

abumalick:
importLoader should be 2 because there is postcss-loader and sass-loader after it.

PS:
I don't understand this option, it works without it, why is it for ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ The query parameter `importLoaders` allows to configure how many loaders before 
     {
       loader: 'css-loader',
       options: {
-        importLoaders: 1 // 0 => no loaders (default); 1 => postcss-loader; 2 => postcss-loader, sass-loader
+        importLoaders: 2 // 0 => no loaders (default); 1 => postcss-loader; 2 => postcss-loader, sass-loader
       }
     },
     'postcss-loader',


evilebottnawi:
@abumalick why your change option in README, if your "I don't understand this option, it works without it, why is it for ?" :smile:

abumalick:
It is not logic 
```
{
  test: /\.css$/,
  use: [
    'style-loader',
    {
      loader: 'css-loader',
      options: {
        importLoaders: 1 // 0 => no loaders (default); 1 => postcss-loader; 2 => postcss-loader, sass-loader
      }
    },
    'postcss-loader',
    'sass-loader'
  ]
}
```
it says: `2 => postcss-loader, sass-loader`
Why we have `importLoaders: 1` here then ? we have `postcss-loader` and `sass-loader` in this example

evilebottnawi:
@abumalick `importLoaders` using for external import in scss, example 
Input (scss):
```scss
@import 'test.css'
@import 'color'; // inside we have `.color { color: red; }`
``` 

Output (`css`):
```css
@import 'test.css'
.color { color: red; }
```
Then `css-loader` handle this import using https://github.com/webpack-contrib/css-loader/blob/master/lib/getImportPrefix.js

Example your have `style-loader` -> `css-loader` -> `postcss-loader` -> `sass-loader`.
Then your have `importLoaders: 1` your got:
```shell
-!/path/to/node_modules/css-loader/index.js?{"import":true,"sourceMap":true,"importLoaders":1}!/path/to/node_modules/postcss-loader/lib/index.js?{"sourceMap":true}!
```
I.e. each `css` import will be handled `css-loader` -> `postcss-loader`

Then your have `importLoaders: 2` your got:
```shell
-!/path/to/node_modules/css-loader/index.js?{"import":true,"sourceMap":true,"importLoaders":2}!/path/to/node_modules/postcss-loader/lib/index.js?{"sourceMap":true}!/path/to/node_modules/sass-loader/lib/loader.js?{"includePaths":["/path/to/src"],"outputStyle":"expanded","precision":8,"sourceMap":true}!
```
I.e. each `css` import will be handled `css-loader` -> `postcss-loader` -> `sass-loader`

If your `css` have `sass`, but it is weird your can use `2` value, if your `css` contain pure css your should use only `postcss-loader` and `importLoaders` should be `1`. Also your can have other loader between `postcss-loader` and `sass-loader`, example `resolve-url` and your want to handle url in improted file - your should set `importLoaders` to `2`.

abumalick:
Ok thank you very much for explanation. Now I understand better. 

Looking for your documentation. 
Thank you for your work

michael-ciniawsky:
Yeah 😛 it test for `.css` files in the example, so `1` is correct here :) (I nearly missed it myself :D). Thx

PR #563, opened 2017-06-25
feat: show warning when `NODE_ENV === 'production'` and `!query.minimize`

jeremenichelli:
Adds a warning on terminal when `NODE_ENV` is `"production"` and `minimize` is _falsy_.

Yes.

No relevant.

**Summary**
While debugging my codebase and seeing the css injected by the `style-loader` after passing through `css-loader`, styles _looked_ minified but they are not, only spaces and line breaks are eliminated. This _wrong impression_ of css being minified when is just being collapsed might be happening to more people, missing the chance to reduce the bundle they are delivering in production.

Recently, Addy Osmani in his recent talk on FluentConf add this as something devs forget to do on production (https://twitter.com/housecor/status/877567271408869376).

No.

**Other information**
I created a separate module to handle warnings, in case somebody wants to extend them in the future. Warning doesn't mutate any object, that's why it receives the process object and the minimize value. This also makes testing super easy.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -7,7 +7,7 @@ var processCss = require("./processCss");
 var getImportPrefix = require("./getImportPrefix");
 var compileExports = require("./compile-exports");
 var createResolver = require("./createResolver");
-
+var warnings = require("./warnings");
 
 module.exports = function(content, map) {
 	if(this.cacheable) this.cacheable();
@@ -18,6 +18,9 @@ module.exports = function(content, map) {
 	var camelCaseKeys = query.camelCase || query.camelcase;
 	var sourceMap = query.sourceMap || false;
 	var resolve = createResolver(query.alias);
+
+  // suggest minimizing to user on production build
+  warnings.minimizeInProduction(query.minimize, process);
 
 	if(sourceMap) {
 		if (map) {
--- a/lib/warnings.js
+++ b/lib/warnings.js
@@ -8,18 +8,15 @@ var cachedProcessId = 0;
  * @param {Object} processObject current process object
  * @returns {undefined}
  */
-function minimizeInProduction(minimize, processObject) {
+function minimizeInProduction(minimize, processObject, loader) {
   if (processObject.env.NODE_ENV === "production" && !minimize && processObject.pid !== cachedProcessId) {
     // update process id
     cachedProcessId = processObject.pid;
 
     // show warning
-    minimizeInProduction.console.warn("[css-loader] Building for production? We suggest enabling minimize (https://github.com/webpack-contrib/css-loader#minimize).\n");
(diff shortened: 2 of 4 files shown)


evilebottnawi:
@jeremenichelli what is use case for this?

jeremenichelli:
@evilebottnawi when users build with `NODE_ENV` set to `production`  and hasn't set `minimize` to `true` they will see something like these.

```sh
$ npm run build
> NODE_ENV="production" webpack

[css-loader] Building for production? We suggest enabling minimize (https://github.com/webpack-contrib/css-loader#minimize).
```

I think it's a useful flag to promote smaller bundles on people who might be copying Webpack configs from boilerplates for example. It's like Webpack warning you about a bundle being to large in a yellow chalk.

lib/loader.js
```diff
@@ -19,6 +19,9 @@ module.exports = function(content, map) {
 	var sourceMap = query.sourceMap || false;
 	var resolve = createResolver(query.alias);
 
+  // suggest minimizing to user on production build
+  warnings.minimizeInProduction(query.minimize, process);
```
```js
if (process.env.NODE_ENV === 'production' && !query.minimize)  {
   this.emitWarning('⚠️ CSS Loader\n\n' + 'Building for production?\n' + 'We suggest enabling minimize (https://github.com/webpack-contrib/css-loader#minimize).\n')
}
```

jeremenichelli: What is `this` here? Does webpack already provides a warning API and I wrote all this code in vain? 😩😆

evilebottnawi: @jeremenichelli yep, `webpack` have api for this purpose (`this` - loader api)

TrySound:
cssnano will be removed in new-loader. Does this make sense?

jeremenichelli:
Quick notes, I'm still using my warnings module for easier testing and to avoid multiple warnings. 

Also, I'm seeing that the team is not completely convinced about this. If you agree there's no value on adding it feel free to close the pull request, no hard feelings at all 😃 

I always prefer submitting a PR than an issue if I'm able to.

michael-ciniawsky:
@TrySound I also have this PR [#5010](https://github.com/webpack/webpack/pull/5010) open to remove this webpack specific behaviour (`minimize`) hopefully altogether in the future, but currently it's nice-to-have.If anyone as objections against it though, we might better leave it :)

d3viant0ne:
To @TrySound's point, given this is in no way applicable to the new loader, it really doesn't make sense.

We also agreed to no land major features in the style loader chain as they will likely change, no longer exist, no longer be necessary which in the end will only serve to make the transition for users more difficult in the end.

d3viant0ne:
And for reference, A feature is not `semver: Patch`

A feature is a  `semver: Minor` unless it isn't backward compatible in which case it's `semver: Major`

PR #580, opened 2017-07-18
Update README.md to add `style-loader` to npm install command

bbodine1:
Style-loader is required to use css-loader.

Documentation

Not required

Yes

**Summary**
Changes readme from 
`npm install --save-dev css-loader`
to
`npm install --save-dev css-loader style-loader`

New users can not make css-loader work without also having style-loader installed.

No. Documentation only.

**Other information**

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 <h2 align="center">Install</h2>
 
 ```bash
-npm install --save-dev css-loader
+npm install --save-dev css-loader style-loader
 ```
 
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
Thx, but `css-loader` doesn't _really_ need `style-loader` to work correctly, e.g in conjunction with `extract-text-webpack-plugin` and [others](https://github.com/webpack-contrib/css-loader#usage), so `style-loader` isn't a hard dependency of `css-loader`

PR #583, opened 2017-07-23
Inherit getLocalIdent from parent loader

### LINKED ISSUE #524: getLocalIdent doesn't apply for composed classes
It's a bug

`getLocalIdent` works for all directly required css file, but doesn't for composed classes

**If the current behavior is a bug, please provide the steps to reproduce.**

Loader in the webpack config
```js
{
  test: /\.css$/,
  loader: 'css-loader',
  options: {
    modules: true,
    importLoaders: 1,
    getLocalIdent: () => 'foo'
  }
}
```

style1.css
```css
.bar {
  composes: baz from './style2.css';
  color: white;
}
```

style2.css
```css
.baz {
  background: red;
}
```

module.js
```js
require('./style1.css') // { 'bar': 'foo _3eS695' }
```

module.js
```js
require('./style1.css') // { 'bar': 'foo foo' }
```

css-loader@0.26.1
webpack@2.4.1

odensc:
Bug fix.

No. (I couldn't figure out how to compose from a css file with the test helpers)

**Summary**
Someone using one of my libraries ran into an issue where getLocalIdent was not being applied to composed classes. (odensc/css-loader-minify-class#1)

Looking into the source code for css-loader, I found that the loader query is stringified sometimes - which causes it to lose the function reference to `getLocalIdent`. This PR fixes that by inheriting the getLocalIdent function from the parent loader if possible. 

Also fixes #524 which was reported on this repo.

No.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -18,6 +18,19 @@ module.exports = function(content, map) {
 	var camelCaseKeys = query.camelCase || query.camelcase;
 	var sourceMap = query.sourceMap || false;
 	var resolve = createResolver(query.alias);
+	// Inherit getLocalIdent from parent loader
+	/* eslint-disable no-underscore-dangle */
+	if(this._module && this._module.issuer && this._module.issuer.loaders) {
+		var filename = require.resolve("..");
+		var parentLoader = this._module.issuer.loaders.filter(function(loader) {
+			return loader.loader === filename;
+		})[0];
+		if(parentLoader && parentLoader.options) {
+			query.getLocalIdent = parentLoader.options.getLocalIdent;
+		}
+	}
+
+	/* eslint-enable no-underscore-dangle */
 
 	if(sourceMap) {
 		if (map) {


odensc:
(not sure what's wrong with AppVeyor)

michael-ciniawsky:
Could you try with 

**webpack.config.js**
```js
{
  loader: 'css-loader'
  options: {
    ident: 'css', // <= with value isn't important
    getLocalIdent: () => {}
  }
}
```

This should prevent losing the reference to `{Function}` when `options` get stringified. Normally `webpack >= v2.2.1` should add the `ident` automatically per default, but since `>= 2.6.0` this seems to be buggy again 😞 . I had countless issues with that in `postcss-loader` 

In case it doesn't work `this._module` is a deprecated Loader API ⚠️  , so the current fix won't work this way

odensc:
The `ident` option does indeed work! I didn't try it because I saw it was
fixed (guess not anymore). Also good to know that `_module` is deprecated.
:)

PR #593, opened 2017-08-16
fix: stricter `@import` tolerance

swernerx:
I accidentally found this issue when using size-limit on an exported library where we kept the import-normalize from https://www.npmjs.com/package/postcss-normalize inside the exported library.

Issue is that Webpack currently throws on the simple statement as it tries to parse it as a normal import.

To fix I added the failing test first, then fixed the regexp which matches the import statements and made the url handling generally more tolerant.

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -42,7 +42,7 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 		}
 
 		if(options.import) {
-			css.walkAtRules(/import/i, function(rule) {
+			css.walkAtRules(/^import$/i, function(rule) {
 				var values = Tokenizer.parseValues(rule.params);
 				var url = values.nodes[0].nodes[0];
 				if(url.type === "url") {
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -45,9 +45,9 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 			css.walkAtRules(/^import$/i, function(rule) {
 				var values = Tokenizer.parseValues(rule.params);
 				var url = values.nodes[0].nodes[0];
-				if(url.type === "url") {
+				if(url && url.type === "url") {
 					url = url.url;
-				} else if(url.type === "string") {
+				} else if(url && url.type === "string") {
 					url = url.value;
 				} else throw rule.error("Unexpected format " + rule.params);
 				if (!url.replace(/\s/g, '').length) {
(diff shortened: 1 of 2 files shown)


lib/processCss.js
```diff
@@ -42,12 +42,12 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 		}
 
 		if(options.import) {
-			css.walkAtRules(/import/i, function(rule) {
+			css.walkAtRules(/^import$/i, function(rule) {
```
michael-ciniawsky: `/^@import$/i`

lib/processCss.js
```diff
@@ -42,12 +42,12 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
...
 				var values = Tokenizer.parseValues(rule.params);
 				var url = values.nodes[0].nodes[0];
-				if(url.type === "url") {
+				if(url && url.type === "url") {
```
michael-ciniawsky: ?

lib/processCss.js
```diff
@@ -42,12 +42,12 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
...
 					url = url.url;
-				} else if(url.type === "string") {
+				} else if(url && url.type === "string") {
```
michael-ciniawsky: ?

lib/processCss.js
evilebottnawi: @michael-ciniawsky `walkAtRules` already walks only in `@something`.

lib/processCss.js
swernerx: I'd like to get the same error handling about malformed URLs for non-existing URLs. This was actually the original error, where I got something like `can't look up "type" of undefined`.

lib/processCss.js
swernerx: The at-sign is not needed here, right.

lib/processCss.js
evilebottnawi: @swernerx can your add tests for this, thanks!

swernerx: This would be a new kind of test which tests for throwing errors. We do not do this currently in CSS loader it seems... at least not in the `importTest.js` file.

evilebottnawi: @swernerx we should add tests or avoid this fix here (move to separate issue or PR)

swernerx: Done

evilebottnawi: @swernerx oh no :smile: your test was right with `@import`, we should tests about `url && url.type === "url"`

swernerx: That's what I did. Without the new handling it should just throw some hard error.

PR #597, opened 2017-08-22
fix: add support for aliases starting with `/` (`options.alias`)

Atinux:

Improvement in alias handling

Of course

**Summary**

Hi :)

Thank you for this great webpack loader. Actually, we are using it a lot into [nuxt.js](https://github.com/nuxt/nuxt.js).

We have an issue with the css-loader used into the vue files, we have a global webpack alias named `~` and it's used a lot to require file this way: `import '~/assets/my-file.vue'`

Actually, we are using `css-loader` with this configuration:
```js
{
    loader: 'css-loader',
    options: {
      minimize: true,
      importLoaders: 1,
      sourceMap: true,
      alias: {
        '/static': join(this.options.srcDir, 'static'),
        '/assets': join(this.options.srcDir, 'assets')
      }
   }
}
```

So our users can do this into their CSS:

```css
.logo {
  background: url('~/static/logo.png');
}
.logo-2 {
  background: url('/logo.png');
}
```

Before this PR, `.logo` background url is transformed to `/static/logo.png` and considered as an url not processed (even if it's starting by an alias), `.logo-2` is not processed and that's perfect.

Aftet this PR, `.logo` is transformed as well and considered to be a relative path (since starting by a known alias) and `.logo-2` is not changed :+1: 

No breaking change

Related issue to nuxt.js repo: https://github.com/nuxt/nuxt.js/issues/1435

### DIFF
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -83,7 +83,7 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 			exports[exportName] = replaceImportsInString(exports[exportName]);
 		});
 
-		function startByAlias(url) {
+		function isAlias(url) {
 			// Handle alias starting by / and root disabled
 			return url !== options.resolve(url)
 		}
@@ -103,7 +103,7 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 					}
 					break;
 				case "url":
-					if (options.url && item.url.replace(/\s/g, '').length && !/^#/.test(item.url) && (startByAlias(item.url) || loaderUtils.isUrlRequest(item.url, options.root))) {
+					if (options.url && item.url.replace(/\s/g, '').length && !/^#/.test(item.url) && (isAlias(item.url) || loaderUtils.isUrlRequest(item.url, options.root))) {
 						// Don't remove quotes around url when contain space
 						if (item.url.indexOf(" ") === -1) {
 							item.stringType = "";
--- a/lib/processCss.js
+++ b/lib/processCss.js
@@ -83,6 +83,11 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 			exports[exportName] = replaceImportsInString(exports[exportName]);
 		});
 
+		function startByAlias(url) {
+			// Handle alias starting by / and root disabled
+			return url !== options.resolve(url)
+		}
+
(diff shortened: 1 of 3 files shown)


Atinux:
Thanks for the fast merge @evilebottnawi 

When can I expect a release on npm?

Atinux:
I believe it could be a minor change or patch since it's only a bug fix :)

yuchonghua:
Very good support, probably when the merger can be released?
/ping @evilebottnawi

meiyuanbo:
probably when the merger can be released?

evilebottnawi:
@michael-ciniawsky @d3viant0ne friendly ping, need review, thanks!

lib/processCss.js
```diff
@@ -83,6 +83,11 @@ var parserPlugin = postcss.plugin("css-loader-parser", function(options) {
 			exports[exportName] = replaceImportsInString(exports[exportName]);
 		});
 
+		function startByAlias(url) {
```
michael-ciniawsky: `startsWithAlias` => `isAlias[ed]`

michael-ciniawsky:
Relabeled to  🏷  `semver: Patch` for now as I don't think this would be a breaking change, but anyone might chime in and correct me if I'm wrong 😛

evilebottnawi:
@michael-ciniawsky oh, yep, it is `patch`, just invalid choice `semver`

appinteractive:
Can someone click on merge please? 👍

evilebottnawi:
@appinteractive need minimum 3 approved votes
@michael-ciniawsky @d3viant0ne friendly ping

HadiChen:
Can someone click on merge please?

lib/processCss.js
Atinux: I pushed a change to rename it to `isAlias`

appinteractive:
Any update on when this could go into the master? Sorry for poking around 😅

yuchonghua:
Feeling @d3viant0ne busy

liambuchanan:
Haven't done much digging yet as this is a sub-dependency for us, but it appears merging has introduced a bug as processCss is not being instantiated with a resolve method here https://github.com/webpack-contrib/css-loader/blob/master/lib/localsLoader.js#L18

PR #610, opened 2017-09-11
Feature  increase specificity

### LINKED ISSUE #608: Is it possible to wrap class names?
Hi there.
There is a situation. I need to integrate react app in existed WP website. But react's style override WP's style.
To avoid overriding styles I want add specific classname to react root element (isolate react app), but in that case css modules class names has less specificity than default, normalized and/or bootstrap styles.
e.g. 
`.specificClassname label` has more specificity than `.Label__label_hash64`.

So I need option to add wrapper classname before each compiled classname like
`.specificClassname .Label__label_hash64`

How could I do that?

IvanKalinin:
Added support of postcss-increase-specificity plugin which allow us to isolate spa-app that integrated in non-spa. Closes #608


TrySound:
You don't need to extend css-loader with this quite specific behavior. You don't need to use root. Increase plugin provides defaults which may cover your  needs.

TrySound:
Currently we do not accept new features with specific cases. Next implementation will be integrated in webpack out of the box via icss. All css-modules should be used via postcss-loader and plugins.

PR #619, opened 2017-10-06
docs(README): supplement the use case of `minimize` (`options.minimize`)

MagicJohnJang:
Actually we can put `minimize` after `css-loader`.

Supplement  readme.

yup.
yup.
**Summary**
I'm searching the usage to put the `minimize` more lightly.

Nope.

**Other information**

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,16 @@ You can also disable or enforce minification with the `minimize` query parameter
   }
 }
 ```
+or like
+```js
+{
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
Thx, but `toString` is already covered under [Usage](https://github.com/webpack-contrib/css-loader#tostring), `minimize` is just an option unrelated to other chained loaders

PR #627, opened 2017-10-31
fix: proper URL escaping and wrapping (`url()`)

### LINKED ISSUE #615: Missing quotes in url() for data URIs returned from another loader
Hi!

**STR:**

Using node 8.5.0:

1. Clone https://github.com/edmorley/css-loader-missing-quotes-testcase
2. `yarn install`
3. `yarn build`
4. Inspect `dist/styles.bundle.css`

**Expected:**

```css
#inline-data-uri {
  background-image: url("data:image/svg+xml,%3C?xml version='1.0'?%3E %3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E %3Crect x='10' y='10' height='100' width='100' style='stroke:%23ff0000; fill: %230000ff'/%3E %3C/svg%3E");
}

#data-uri-from-loader {
  background-image: url("data:image/svg+xml,%3C?xml version='1.0'?%3E %3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E %3Crect x='10' y='10' height='100' width='100' style='stroke:%23ff0000; fill: %230000ff'/%3E %3C/svg%3E");
}
```

**Actual**

```css
#inline-data-uri {
  background-image: url("data:image/svg+xml,%3C?xml version='1.0'?%3E %3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E %3Crect x='10' y='10' height='100' width='100' style='stroke:%23ff0000; fill: %230000ff'/%3E %3C/svg%3E");
}

#data-uri-from-loader {
  background-image: url(data:image/svg+xml,%3C?xml version='1.0'?%3E %3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E %3Crect x='10' y='10' height='100' width='100' style='stroke:%23ff0000; fill: %230000ff'/%3E %3C/svg%3E);
}
```

ie: Including the data URI inline in the CSS file works fine (thanks to the fix in #15), however when the resource is `url('./svg-data-uri.txt')` and so processed by another loader (which for simplicity for this testcase is just the `raw-loader`) the data URI string returned by the loader isn't wrapped in quotes, so is invalid CSS.

Contrast this to html-loader by inspecting `dist/index.html`, which successfully converts:
```html
<img src="./svg-data-uri.txt">
```

...to:
```html
<img src="data:image/svg+xml,%3C?xml version='1.0'?%3E %3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E %3Crect x='10' y='10' height='100' width='100' style='stroke:%23ff0000; fill: %230000ff'/%3E %3C/svg%3E">
```

Due to this issue, [svg-url-loader](https://github.com/bhovhannes/svg-url-loader) unfortunately defaults to adding extra quotes around the string it returns, so that CSS `url()` references work. However this then breaks the `html-loader` (bhovhannes/svg-url-loader#126) and "require() string from JS" use-cases, which require consumers to remember to set the loader's `noquotes` option to `false` (eg mozilla-neutrino/neutrino-dev#334).

However I think this is ideally something that should be fixed in `css-loader` itself, meaning `svg-url-loader` could stop wrapping everything in quotes, solving the other problems.

Thoughts? :-)

For quick reference, the relevant package versions in the testcase repo's `yarn.lock` are:
_webpack 3.6.0, css-loader 0.28.7, raw-loader 0.5.1, html-loader 0.5.1, extract-text-webpack-plugin 3.3.0._

kgram:
Bugfix

Yes. 2 existing tests changed to account for new behaviour (no single-quotes), 6 new tests of escaping.

I don't believe it's relevant. Unless just to document the quoting behaviour.

**Summary**
Fixes #615. Previously, modules had their quotes stripped unless the module-path contained a space. Now, quotes are always stripped and the URL returned from the module is escaped and wrapped in double-quotes if necessary. Modules returning a quote-wrapped string will have their own wrapping-quotes stripped before wrapping/escaping.

I don't think it does, modules wrapping in quotes like svg-url-loader should still function.

The only consequence is that it is impossible to wrap in single-quotes, although it's hard to see when this would be a necessity.

**Other information**
There is one weird case I'm not sure how to handle: newlines should be escaped in the quoted url. [The spec states](https://www.w3.org/TR/css-syntax/#newline) that only line feeds (`\n`) count as newlines, since carriage return (`\r`) and form feed (`\f`) are automatically converted during preprocessing. However, if left unescaped they would be converted to unescaped line feeds. If escaped, they wouldn't be converted. I'm not sure if the best step would be to convert them to escaped line feeds manually or if it even matters. Thoughts?

### DIFF
--- a/lib/escape-url.js
+++ b/lib/escape-url.js
@@ -1,0 +1,13 @@
+module.exports = function(url) {
+    // If url is already wrapped in quotes, remove them
+    if ((url[0] === '"' || url[0] === '\'') && url[0] === url[url.length - 1]) {
+        url = url.slice(1, -1);
+    }
+    // Should url be wrapped?
+    // See https://drafts.csswg.org/css-values-3/#urls
+    if (/["'() \t\n]/.test(url)) {
+        return '"' + url.replace(/"/g, '\\"').replace(/\n/g, '\\n') + '"'
+    }
+
+    return url
+}
--- a/lib/escape-url.js
+++ b/lib/escape-url.js
@@ -1,6 +1,6 @@
 module.exports = function(url) {
     // If url is already wrapped in quotes, remove them
-    if ((url[0] === '"' || url[0] === '\'') && url[0] === url[url.length - 1]) {
+    if (/^['"].*['"]$/.test(url)) {
         url = url.slice(1, -1);
     }
     // Should url be wrapped?
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -82,10 +82,13 @@ module.exports = function(content, map) {
 				"[" + JSON.stringify(importItem.export) + "] + \"";
 		}
 
-		var escapeUrlHelper;
 		cssAsString = cssAsString.replace(result.importItemRegExpG, importItemMatcher.bind(this));
+
+		// helper for ensuring valid CSS strings from requires
+		var urlEscapeHelper = "";
+
 		if(query.url !== false && result.urlItems.length > 0) {
-			escapeUrlHelper = "var escapeUrl = require(" +
+			urlEscapeHelper = "var escape = require(" +
 				loaderUtils.stringifyRequest(this, require.resolve("./escape-url.js"))
 				+ ")\n"
 
@@ -100,14 +103,12 @@ module.exports = function(content, map) {
 				if(idx > 0) { // idx === 0 is catched by isUrlRequest
 					// in cases like url('webfont.eot?#iefix')
 					urlRequest = url.substr(0, idx);
-					return "\" + escapeUrl(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"" +
+					return "\" + escape(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"" +
 							url.substr(idx);
 				}
 				urlRequest = url;
-				return "\" + escapeUrl(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"";
+				return "\" + escape(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"";
 			}.bind(this));
-		} else {
-			escapeUrlHelper = ""
 		}
 		
 		var exportJs = compileExports(result, importItemMatcher.bind(this), camelCaseKeys);
@@ -133,7 +134,7 @@ module.exports = function(content, map) {
 		}
 
 		// embed runtime
-		callback(null, escapeUrlHelper +
+		callback(null, urlEscapeHelper +
 			"exports = module.exports = require(" +
 			loaderUtils.stringifyRequest(this, require.resolve("./css-base.js")) +
 			")(" + query.sourceMap + ");\n" +
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -88,9 +88,7 @@ module.exports = function(content, map) {
 		var urlEscapeHelper = "";
 
 		if(query.url !== false && result.urlItems.length > 0) {
-			urlEscapeHelper = "var escape = require(" +
-				loaderUtils.stringifyRequest(this, require.resolve("./escape-url.js"))
-				+ ")\n"
+			urlEscapeHelper = "var escape = require(" + loaderUtils.stringifyRequest(this, require.resolve("./url/escape.js")) + ")\n";
 
 			cssAsString = cssAsString.replace(result.urlItemRegExpG, function(item) {
 				var match = result.urlItemRegExp.exec(item);
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -82,8 +82,13 @@ module.exports = function(content, map) {
 				"[" + JSON.stringify(importItem.export) + "] + \"";
 		}
 
+		var escapeUrlHelper;
 		cssAsString = cssAsString.replace(result.importItemRegExpG, importItemMatcher.bind(this));
-		if(query.url !== false) {
+		if(query.url !== false && result.urlItems.length > 0) {
+			escapeUrlHelper = "var escapeUrl = require(" +
+				loaderUtils.stringifyRequest(this, require.resolve("./escape-url.js"))
+				+ ")\n"
+
 			cssAsString = cssAsString.replace(result.urlItemRegExpG, function(item) {
 				var match = result.urlItemRegExp.exec(item);
 				var idx = +match[1];
@@ -95,15 +100,16 @@ module.exports = function(content, map) {
 				if(idx > 0) { // idx === 0 is catched by isUrlRequest
 					// in cases like url('webfont.eot?#iefix')
 					urlRequest = url.substr(0, idx);
-					return "\" + require(" + loaderUtils.stringifyRequest(this, urlRequest) + ") + \"" +
+					return "\" + escapeUrl(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"" +
 							url.substr(idx);
 				}
 				urlRequest = url;
-				return "\" + require(" + loaderUtils.stringifyRequest(this, urlRequest) + ") + \"";
+				return "\" + escapeUrl(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"";
 			}.bind(this));
+		} else {
+			escapeUrlHelper = ""
 		}
-
-
+		
 		var exportJs = compileExports(result, importItemMatcher.bind(this), camelCaseKeys);
 		if (exportJs) {
 			exportJs = "exports.locals = " + exportJs + ";";
@@ -127,7 +133,8 @@ module.exports = function(content, map) {
 		}
 
 		// embed runtime
-		callback(null, "exports = module.exports = require(" +
+		callback(null, escapeUrlHelper +
+			"exports = module.exports = require(" +
 			loaderUtils.stringifyRequest(this, require.resolve("./css-base.js")) +
 			")(" + query.sourceMap + ");\n" +
 			"// imports\n" +
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -88,7 +88,7 @@ module.exports = function(content, map) {
 		var urlEscapeHelper = "";
 
 		if(query.url !== false && result.urlItems.length > 0) {
-			urlEscapeHelper = "var escape = require(" + loaderUtils.stringifyRequest(this, require.resolve("./url/escape.js")) + ")\n";
+			urlEscapeHelper = "var escape = require(" + loaderUtils.stringifyRequest(this, require.resolve("./url/escape.js")) + ");\n";
 
 			cssAsString = cssAsString.replace(result.urlItemRegExpG, function(item) {
 				var match = result.urlItemRegExp.exec(item);
--- a/test/moduleTestCases/urls/expected.css
+++ b/test/moduleTestCases/urls/expected.css
@@ -2,7 +2,7 @@
 	background: url({./module});
 	background: url({./module});
 	background: url("{./module module}");
-	background: url('{./module module}');
+	background: url("{./module module}");
 	background: url({./module});
 	background: url({./module}#?iefix);
 	background: url("#hash");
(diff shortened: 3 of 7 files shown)


lib/escape-url.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function(url) {
+    // If url is already wrapped in quotes, remove them
+    if ((url[0] === '"' || url[0] === '\'') && url[0] === url[url.length - 1]) {
```
michael-ciniawsky:
```js
if (/^['"].*['"]$/.test(url)) {
  url = url.slice(1, -1);
}
```

michael-ciniawsky:
```js
const url1 = 'path/to/image.png'
const url2 = '"path/to/image.png"'

console.log(JSON.stringify(url1))
console.log(JSON.stringify(url2))
```

```
"path/to/image.png"
"\"path/to/image.png\""
```

michael-ciniawsky:
Could you check if 

```js
return JSON.stringify(require(" + loaderUtils.stringifyRequest(this, urlRequest) + "));
```

also works ? Not 💯  if I placed it correctly :)

kgram:
I can't find a reason `JSON.stringify` shouldn't work, there doesn't seem to be any conflicts between the formats. It _feels_ a little wrong, though, which I guess is why I spent so long looking for a problem. But I guess if it works that ought to be enough.

Two things to note: if you replace the entire `escapeUrl` function with `JSON.stringify` all URLs are wrapped in quotes no matter their content and quote-wrapped strings (such as from `svg-url-loader`) will keep their quotes. I think it would have to be released as a major version to avoid breaking changes then. Maybe an inline `.replace(/^['"](.*)['"]$/, '$1')` could be used to avoid it.

Your call :)

lib/escape-url.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function(url) {
+    // If url is already wrapped in quotes, remove them
+    if (/^['"].*['"]$/.test(url)) {
+        url = url.slice(1, -1);
+    }
+    // Should url be wrapped?
+    // See https://drafts.csswg.org/css-values-3/#urls
+    if (/["'() \t\n]/.test(url)) {
```
michael-ciniawsky: What exactly is tested here, could you give a few examples please ?

1. `["'()=> <=\t\n]` => `\s` ?
2. `/["'() \t\n]/.test(url)` => `/["'() \t\n]/g.test(url)` (`/RegExp/g`) ?

lib/loader.js
```diff
@@ -82,8 +82,13 @@ module.exports = function(content, map) {
 				"[" + JSON.stringify(importItem.export) + "] + \"";
 		}
 
+		var escapeUrlHelper;
```
```js
// url() helper to escape quotes
var escape = ""
```

`\n` after the variable decl please

lib/loader.js
```diff
@@ -82,8 +82,13 @@ module.exports = function(content, map) {
...
 		cssAsString = cssAsString.replace(result.importItemRegExpG, importItemMatcher.bind(this));
-		if(query.url !== false) {
+		if(query.url !== false && result.urlItems.length > 0) {
+			escapeUrlHelper = "var escapeUrl = require(" +
```
```js
escape = "require(" + loaderUtils.stringifyRequest(this, require.resolve("./url/escape.js")) + ")\n"
```

lib/loader.js
```diff
@@ -95,15 +100,16 @@ module.exports = function(content, map) {
...
 					urlRequest = url.substr(0, idx);
-					return "\" + require(" + loaderUtils.stringifyRequest(this, urlRequest) + ") + \"" +
+					return "\" + escapeUrl(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"" +
 							url.substr(idx);
 				}
 				urlRequest = url;
-				return "\" + require(" + loaderUtils.stringifyRequest(this, urlRequest) + ") + \"";
+				return "\" + escapeUrl(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"";
 			}.bind(this));
+		} else {
```
michael-ciniawsky: rm the `else` as initializing `escape` with a default value (`""`) should be enough

test/urlTest.js
```diff
@@ -103,6 +103,25 @@ describe("url", function() {
...
+	], "", { './module': "module\"with\"quote" });
+	test("module with quote wrapped", ".class { background: green url(module) xyz }", [
+		[1, ".class { background: green url(\"module\\\"with\\\"quote\\\"wrapped\") xyz }", ""]
+	], "", { './module': "\"module\"with\"quote\"wrapped\"" });
+	test("module with parens", ".class { background: green url(module) xyz }", [
+		[1, ".class { background: green url(\"module(with-parens)\") xyz }", ""]
+	], "", { './module': 'module(with-parens)' });
+	test("module with newline", ".class { background: green url(module) xyz }", [
+		[1, ".class { background: green url(\"module\\nwith\\nnewline\") xyz }", ""]
+	], "", { './module': "module\nwith\nnewline" });
```
michael-ciniawsky: We need to test `url-loader` output here aswell

lib/escape-url.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function(url) {
```
michael-ciniawsky: `lib/escape-url-js` => `lib/url/escape.js` 

`module.exports = function escape (url) {`

So we hopefully can require the named `{Function}` without having to decl a new variable later :)

michael-ciniawsky:
What about data URL's, e.g from the `url-loader` ?

lib/escape-url.js
kgram: They are the characters outlined as requiring escaping in unquoted URLs in the CSS spec.

https://drafts.csswg.org/css-values-3/#urls

If any of them are present, the string is wrapped in quotes and quotes/newline is escaped. Like with `JSON.stringify`, yes, it's basically the same as `\s` in regex, just from different standards. There seems to be some rarely used characters regex considers whitespace that CSS doesn't, but again, it is unlikely to matter. I just prefer to be specific when the standard is right there.

Some examples that would require either escaping or wrapping:

`url(open(parens)` => `url(open\(parens)` / `url("open(parens")`

`url(close)parens)` => `url(close\)parens)` / `url("close)parens")`

`url(quote"in"middle)` => `url("quote\"in\"middle")`

lib/loader.js
kgram: I'm a little confused here. Surely you mean `escape = "var escape = require(" + ...`, right?

Don't you think it gets a little confusing when you call both the variable in the loader and in the string by the same name?

lib/escape-url.js
kgram: Done.

lib/escape-url.js
kgram: Done.

lib/escape-url.js
kgram: Regarding your second point: Why would it be better to test with a global regex? I'm not aware of any difference caused by this.

lib/loader.js
kgram: The variable is now named `urlEscapeHelper` in the loader and `escape` in the output. If you feel that is incorrect could you be a bit more specific about the naming you would prefer?

lib/loader.js
kgram: Done.

test/urlTest.js
kgram: Test added of one of the url-loader outputs from [this article](https://julienrenaux.fr/2015/03/30/introduction-to-webpack-with-practical-examples/). It's a very small file, but as far as I know all base64-encoded data-URIs are valid in CSS when unquoted. If you have an example in mind of something that might fail I'll add that too though.

kgram:
I think everything is either fixed or addressed now :)

evilebottnawi:
Just information, https://github.com/webpack-contrib/sass-loader/issues/490 maybe related

kgram:
@michael-ciniawsky, @evilebottnawi: Have you seen my comments? I think I've either fixed or addressed everything now, could you have a look at it again?

test/moduleTestCases/urls/expected.css
```diff
@@ -2,7 +2,7 @@
 	background: url({./module});
 	background: url({./module});
 	background: url("{./module module}");
-	background: url('{./module module}');
+	background: url("{./module module}");
```
evilebottnawi: Why we change `'` to `"`? Can we save original quotes?

kgram: In principle it could be done. It will complicate things quite a bit, though, since the whole point of the update is to quote based on module-content instead of name. The type of quotes would have to be passed to the client (something like `escapeSingle(require(<resource>))`).

I skipped it, since I can't really imagine a scenario where it would be important to use single quotes. `svg-url-loader` already assumes double-quotes, since this is what is commonly used in HTML, I assume other loaders will make similar choices. If not, the worst case scenario is a slightly longer string.

If this is implemented, there should maybe also be an `escapeUnquoted` function for handling unquoted modules. That way the quotes used will be entirely up to the user, the loader will just ensure valid CSS is produced.

d3viant0ne: `I assume other loaders will make similar choices` - If there is one thing I have learned about this ecosystem, that assumption is almost guaranteed to break something.

kgram: Right, but "break something" here means "result in a slightly longer string since double-quotes are escaped". The whole point of this change is to make it impossible for other loaders to break your CSS.

My arguments for using only double-quotes summed up:

1. Double-quote wrapping is the expected default for URLs in HTML in all the view-engines I'm familiar with, so I think it is a fair assumption that most loaders will expect this and avoid double-quotes. If they don't, nothing breaks.
2. A user shouldn't have to consider this detail. My own sass-lint rules force single-quotes since this is used across the project, which would work against `svg-url-loader`'s change from double to single-quotes.
3. I don't think the quotes written by the user ought to be considered equal to the quotes produced in the output. The user quotes the module-name, the loader quotes the actual URL, there's no reason to expect similar requirements.

Of course it is not my package, and I'm not that experienced with writing loaders, so I defer entirely to your judgements. If I have failed to convince you, I'll change it so there is a separate escape-function for single-quotes. If that is the case, should I also make an unquoted escape-function?

d3viant0ne: I'm no debating `correctness` I'm just saying that I can all but guarantee that at some point after this lands in master, it will have broken something, somewhere. 

That has nothing to do with not merging it, just saying it's going to happen & we now we have context as to why.

d3viant0ne: Try to keep in mind that `css-loader` sits at the core of the entire style chain and in some places, that's a long & convoluted chain for better or worse.

Anything that lands in this repo has to be vetted, pondered, picked apart & poked so we at the very least can have a reason why there are 4 new issues with version `0.0.x` broke `Y`

Changes in behaviors get that & the rubber glove treatment, it's just too easy to not account for a path in this vast ecosystem where Change X has an unintended side effect on Y  & people show up with torches & pitch forks.

evilebottnawi:
@kgram friendly ping

kgram:
Are you waiting for something from me? I was waiting for you to finish reviewing. I guess that explains why nothing was happening :)

I think I've responded to every comment so far. Is there something specific I need to do, or have I missed something?

kgram:
@evilebottnawi: given your lack of response, I'm starting to suspect you aren't subscribed to updates from the thread. If that is the case, I apologise for not @-mentioning you in my comments. I just assumed all reviewers would subscribe to the thread, but I can see how that could be inconvenient when managing large projects. At any rate, to repeat my answer to your ping:

Are you waiting for something from me? I was waiting for you to finish reviewing. I guess that explains why nothing was happening :)

I think I've responded to every comment so far. Is there something specific I need to do, or have I missed something?

evilebottnawi:
@michael-ciniawsky @d3viant0ne friendly ping

DawnWright:
This PR introduced the following bug: https://github.com/webpack-contrib/css-loader/issues/659

PR #655, opened 2018-01-05
feat(plugins/url): add `url()` filter support (`options.url`)

### LINKED ISSUE #455: Option to ignore not found modules/URLs
It would be good to have an option to skip not found URLs instead of throwing "Module not found" errors.
It might be just a warning instead of critical error, or just skipped.
Ignoring all URLs and @imports, which is already supported - is not an option, since it's too global.

When URL is not found it throws the error in following format:
```
ERROR in ./~/css-loader!./src/css/site.css
Module not found: Error: Can't resolve '../../node_modules/somelibrary/not-existing.gif'
```

This error eventually fails the whole webpack build.

If some options is enabled, such error becomes not critical and build is not marked as failed.

Motivation is that when you rely on third-party library you cannot be sure they will include all files referenced in their `vendor.css`. We are forced (corporate branding stuff) to rely on some internal (but still third-party library - i.e. different team responsibility), which works following some strange bureaucracy workflow and bad publishing practices. So not including all referenced files - is a common case..

Would it be possible to implement such feature or are there any blockers why it wasn't made already so?
Or maybe there are other ways to achieve same thing?

### LINKED ISSUE #538: How to ignore specific url
Hello. I want to know how to ignore specific url.
Using option `url: false` doesn't help, because it disable resolving of all `url` usage. Also the usage of `/` doesn't help.
Is it possible enable ignoring for specific path? Like below:
```scss
.some-class {
  background: url("../../specific-folder/some-image.png") // want to output url with specific-folder as is
}
```

Thanks.

### LINKED ISSUE #558: Rewrite `@import` and `url()` resolution as PostCSS plugins
I'm a member of the cssnano team. Due to the way css-loader works, we've had to make certain allowances to the code in order to keep everything functioning correctly \[[1](https://github.com/ben-eb/cssnano/blob/master/packages/postcss-ordered-values/src/index.js#L33)]\[[2](https://github.com/ben-eb/cssnano/blob/master/packages/postcss-minify-params/src/index.js#L32-L38)]. Unfortunately, as the code keep on evolving, [we continue running into issues](https://github.com/ben-eb/cssnano/issues/376#issuecomment-309776246) which has to do with the integration in css-loader. Rather than continuing the trend of implementing hacks in individual packages, we'd rather fix the issue at its source. 

Upon reviewing css-loader's code base, we learned that cssnano is included in a postcss pipeline preceded by a local plugin called `css-loader-parser` which replaces `@import` and `url()` references with `___CSS_LOADER_` constants. This causes some issues for us because at the time the CSS AST reaches cssnano, the CSS is still unfinished and is actually finalised using regex replacement after the pipeline has finished. Not only is this problematic for us, but it [may also cause issues when control is returned back to webpack](https://github.com/postcss/postcss-loader/issues/262#issuecomment-309204871) due to the CSS (having been minified and optimised) no longer is what css-loader expects (which is seemingly also why css-loader [excludes certain optimisations](https://github.com/webpack-contrib/css-loader/issues/265)). Rather than working on unfinished CSS and leaving users with a diminished experience, we feel that cssnano should be invoked at the very end, working with the fully processed CSS. 

After some discussion between me and @ben-eb, we concluded that the ideal course of action would be to remove the regex replacements and instead replace them with postcss plugins, allowing all processing to be done within the pipeline. This would not only solve the issues for us, but would probably also make css-loader a bit more performant and manageable as well. 

We'd love to hear your thoughts on this.

### LINKED ISSUE #604: Prevent css-loader from requiring urls but also prepend root
I have my folder structure like so:

```
myApp
 +-- src
 |     |-- index.js
 |     |-- index.css
 +-- resources
 |     |-- fonts
 |     +-- images
 |     |     |-- portrait.png
 |     |     |-- bg.png
```

My webpack I have this 
```
rules: [
    {
        test: /\.css$/,
        use: ExtractTextPlugin.extract({
            fallback: 'style-loader',
            use: { loader: 'css-loader', options: { url: true, root: 'http://localhost:8080/' } }
        }),
        exclude: /node_modules/
    },
    {
        test: /(\.png|\.jpg|\.otf)$/,
        use: ['file-loader?name=[name].[ext]&publicPath=assets/&outputPath=assets/']
    }
]
```

In the index.js file I have this

```
import './resources/images/portrait.png' // <- this is handled by the file loader
import './index.css' // <- this should be handled by the ExtractTextPlugin and css-loader
// blah blah
export default Index;
```

In the index.css file I have this:

```
.background {
    background: url('/assets/portrait.png') 0 center;
}
```

So I have to have `{ url: true }` in order for the `{ root: 'http://localhost:8080/' }` to be prepended - however, it seems like it's trying to make a js module and requiring in the asset, which isn't what I want.....

Is there anyway to prevent the css-loader from trying to resolve the url but still have the root prepended

### LINKED ISSUE #630: Flexible url resolving behavior
feature

urls are resolved or not. one global flag via config.

allow some urls to pass unresolved:
url(media/image.jpg')
url(./media/image.jpg')
url(/media/image.jpg')

some prefix would be great:
url(!/media/image.jpg')

can't make specific urls that will respect header 'base' meta. root urls do not respect 'base' meta at all.
in case if app hosted in subdirectory like domain.com/forum

michael-ciniawsky:
### `Issues`

- Fixes #455 
- Fixes #538
- Fixes #558
- Fixes #604
- Fixes #630
 
#### `{Boolean}` (default: `true`)

**webpack.config.js**
```js
{
  loader: 'css-loader',
  options: {
    url: false
  }
}
```

#### `{RegExp}`

**webpack.config.js**
```js
{
  loader: 'css-loader',
  options: {
    url: /filter/
  }
}
```

#### `{Function}`

**webpack.config.js**
```js
{
  loader: 'css-loader',
  options: {
    url (url) {
      return /filter/.test(url)
    }
  }
}
```


README.md
```diff
@@ -88,6 +89,32 @@ To disable `url()` resolving by `css-loader` set the option to `false`
 }
 ```
 
+#### `{RegExp}`
```
stereokai: A few words about what these options do would be smashing!

michael-ciniawsky: yeah yeah :) I intend to appropriate documentation before this goes anywhere, it hasn't even been signed-off by other maintainers nor released yet... 😛

stereokai: Oh man, was already hoping to use it today 😆 😭

bmfs:
What is the hold on this, or what are the alternatives for achieving this?

SergeyKharchenko:
@bmfs +100

SergeyKharchenko:
@stereokai, when can we expect this publicly?

unformatt:
+1

randalpinto:
+1

alekaizer:
+1

slint:
+1

PR #663, opened 2018-01-16
fix: ignore invalid URLs (`url()`)

### LINKED ISSUE #659: v0.28.8 regression: Relative urls in @font-face throw exception `TypeError: url.replace is not a function`
This bug was introduced in v0.28.8, and does not occur in v0.28.7 - it was introduced last week by https://github.com/webpack-contrib/css-loader/pull/627/

Repro:
Use a relative url inside a css `@font-face`, e.g.
```
@font-face {
    font-family: 'Test Font';
    src: url('../MyFont.woff2');
}
```

RESULT: `css-loader` throws exception `TypeError: url.replace is not a function`:
```
Module build failed: TypeError: url.replace is not a function
    at escape (webpack-internal:///1:11:26)
    at eval (webpack-internal:///0:7:71)
    at Object.<anonymous> (/redactedpath/node_modules/css-loader/index.js??ref--1-1!/redactedpath/node_modules/sass-loader/lib/loader.js!/redactedpath/node_modules/stripcomment-loader/index.js!/redactedpath/app/css/styles.scss:71:1)
    at __webpack_require__ (/redactedpath/node_modules/css-loader/index.js??ref--1-1!/redactedpath/node_modules/sass-loader/lib/loader.js!/redactedpath/node_modules/stripcomment-loader/index.js!/redactedpath/app/css/styles.scss:21:30)
    at /redactedpath/node_modules/css-loader/index.js??ref--1-1!/redactedpath/node_modules/sass-loader/lib/loader.js!/redactedpath/node_modules/stripcomment-loader/index.js!/redactedpath/app/css/styles.scss:64:18
    at Object.<anonymous> (/redactedpath/node_modules/css-loader/index.js??ref--1-1!/redactedpath/node_modules/sass-loader/lib/loader.js!/redactedpath/node_modules/stripcomment-loader/index.js!/redactedpath/app/css/styles.scss:67:10)
    at Module._compile (module.js:624:30)
    at Object.exec (/redactedpath/node_modules/webpack/lib/NormalModule.js:129:12)
    at /redactedpath/node_modules/extract-text-webpack-plugin/dist/loader.js:131:26
    at compile (/redactedpath/node_modules/webpack/lib/Compiler.js:300:11)```
```

Note, the URL is indeed a valid path, and I've confirmed that I get the appropriate `ModuleNotFoundError` with invalid relative paths. This bug only occurs when the relative path points to a valid file location.

Also note, the bug does not occur with an absolute path to the same file. Unfortunately, using absolute path is not a feasible workaround since the css in question is an external file we have no control over.

My understanding is that a relative url tells the css-loader to ignore the url. `resolve` therefore returns `{}` as the URL, which is passed to the new `escape` function and thus throws an exception calling `replace`:
```
module.exports = function escape(url) {
    // If url is already wrapped in quotes, remove them
    if (/^['"].*['"]$/.test(url)) {
        url = url.slice(1, -1);
    }
    // Should url be wrapped?
    // See https://drafts.csswg.org/css-values-3/#urls
    if (/["'() \t\n]/.test(url)) {
        return '"' + url.replace(/"/g, '\\"').replace(/\n/g, '\\n') + '"'
    }

    return url
}
```

We have reverted to v0.28.7 temporarily to avoid this bug.

kgram:

Closes #659

Test added for dependency exporting an empty object and null.

Not relevant

**Summary**

https://github.com/webpack-contrib/css-loader/pull/627 introduced an assumption about all dependencies exporting strings. To avoid including font-files, some users use `null-loader`, which effectively exports an empty object.

This PR checks for non-strings, restoring previous functionality for all non-string dependencies.

While this PR would enable previous usage patterns, I hope someone can come up with a better solution. The users are relying on `url([object Object])` being invalid CSS to avoid making requests for font files. This seems fragile in the long run.

Ideally, a null/undefined value (or what we here know *should* be a null value, `{}`) should result in no request being made. As far as I know, this requires the entire `url(...)` block be replaced with `none`, and I'm still not sure this would be valid at all times. This would be a complicated solution.

No.

### DIFF
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -16,7 +16,7 @@ function getEvaluated(output, modules) {
 				return require("../lib/url/escape");
 			if(module.indexOf("-!/path/css-loader!") === 0)
 				module = module.substr(19);
-			if(modules && modules[module])
+			if(modules && module in modules)
 				return modules[module];
 			return "{" + module + "}";
 		});
--- a/lib/url/escape.js
+++ b/lib/url/escape.js
@@ -1,4 +1,7 @@
 module.exports = function escape(url) {
+    if (typeof url !== 'string') {
+        return url
+    }
     // If url is already wrapped in quotes, remove them
     if (/^['"].*['"]$/.test(url)) {
         url = url.slice(1, -1);
--- a/test/urlTest.js
+++ b/test/urlTest.js
@@ -125,6 +125,12 @@ describe("url", function() {
 	test("module from url-loader", ".class { background: green url(module) xyz }", [
 		[1, ".class { background: green url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAA) xyz }", ""]
 	], "", { './module': "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAA" });
+	test("module from null-loader (empty object from webpack)", ".class { background: green url(module) xyz }", [
+		[1, ".class { background: green url([object Object]) xyz }", ""]
+	], "", { './module': {} });
+	test("module is null", ".class { background: green url(module) xyz }", [
+		[1, ".class { background: green url(null) xyz }", ""]
+	], "", { './module': null });
 
 	test("background img with url", ".class { background: green url( \"img.png\" ) xyz }", [
 		[1, ".class { background: green url( \"img.png\" ) xyz }", ""]


test/helpers.js
```diff
@@ -16,7 +16,7 @@ function getEvaluated(output, modules) {
 				return require("../lib/url/escape");
 			if(module.indexOf("-!/path/css-loader!") === 0)
 				module = module.substr(19);
-			if(modules && modules[module])
+			if(modules && module in modules)
```
evilebottnawi: Why `in`?

kgram: @evilebottnawi: To allow for falsy modules such as `{ './module': null }` to still be resolved.

evilebottnawi: @kgram thanks for clafiry :+1:

kgram:
@evilebottnawi: Do you know what is wrong with the travis-build? It looks like the processes were aborted halfway through when looking at the logs.

evilebottnawi:
@kgram do not pay attention (legacy), main build doing in `ci/circleci`

michael-ciniawsky:
Released in [v0.28.9](https://github.com/webpack-contrib/css-loader/releases/tag/v0.28.9) 🎉

Issue #676, opened 2018-02-13
css-loader fails to bundle bootstrap.min.css in Server-side rendering, while sass-loader succeeds


BurkovBA:
I'm trying to create a server-side rendering configuration for my React app with Webpack and Express.js and it almost works.

I managed to bundle my `.jsx` and `sass`, not using `style-loader` (which is client-only), but going with `ExtractTextPlugin` instead. 

All was good, but then I tried to include css files into the bundle and on any css file, e.g. `bootstrap.min.css`, webpack chokes, although my SASS works fine. Here's the error message:

```
$ node_modules/webpack/bin/webpack.js --config ssr-webpack.config.js

Hash: 3e273674716344c7aa0b
Version: webpack 2.7.0
Time: 7982ms
     Asset     Size  Chunks                    Chunk Names
 server.js  3.82 MB       0  [emitted]  [big]  main
server.css  75.1 kB       0  [emitted]         main
   [4] ./src/services/translate.jsx 1.81 kB {0} [built]
   [6] ./src/pages/Blog.jsx 19.1 kB {0} [built]
   [7] ./src/pages/Post.jsx 10.5 kB {0} [built]
   [8] ./src/pages/Layout.jsx 13 kB {0} [built]
  [22] ./src/pages/Home.jsx 2.96 kB {0} [built]
  [27] ./src/pages/posts/2017-10-21-1/index.jsx 9.71 kB {0} [optional] [built]
  [28] ./src/pages/posts/2017-10-25-1/index.jsx 12.7 kB {0} [optional] [built]
  [29] ./src/pages/posts/2017-11-01-1/index.jsx 21 kB {0} [optional] [built]
  [30] ./src/pages/posts/2017-11-21-1/index.jsx 29.3 kB {0} [optional] [built]
  [31] ./src/pages/posts/2017-11-30-1/index.jsx 24.2 kB {0} [optional] [built]
  [39] ./src/server/app.jsx 4.02 kB {0} [built]
  [41] ./src/pages/Blog.scss 41 bytes {0} [built]
  [42] ./src/pages/Post.scss 41 bytes {0} [built]
  [43] ./src/styles/style.scss 41 bytes {0} [built]
  [57] ./src/pages/posts \.jsx 624 bytes {0} [built]
    + 49 hidden modules
Child extract-text-webpack-plugin:
       [0] ./~/css-loader/lib/css-base.js 2.26 kB {0} [built]
[...]

PR #677, opened 2018-02-14
docs(README): add server-side rendering (SSR) example

BurkovBA:
Added documentation of server-side rendering `css-loader` setup to README.md

Not applicable

Yes

**Summary**
This PR is [based on my experience](https://github.com/webpack-contrib/css-loader/issues/676#issuecomment-365589522), building a bundle for server-side rendering.

No

**Other information**

### DIFF
--- a/README.md
+++ b/README.md
@@ -493,21 +493,16 @@ module.exports = {
 
 ### Server-side rendering
 
-Suppose that you want to create a webpack build of your [isomorphic](https://www.lullabot.com/articles/what-is-an-isomorphic-application)/[universal](http://www.acuriousanimal.com/2016/08/10/universal-applications.html) web application that pre-renders pages on the server side. Your app is written in something like ECMA6 and SASS and requires transpilation, and also has CSS dependencies, such as `bootstrap.min.css`, installed from `node_modules`.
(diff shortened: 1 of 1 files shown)


BurkovBA:
@michael-ciniawsky 

Hi, Michael, thanks for review. Yes, I'm not 100% sure about the scope, too. 

On the other hand, it might make sense to provide an example of using `css-loader` in server-side environment.

Maybe, It is too verbose now? I'll try to shorten it.

evilebottnawi:
@michael-ciniawsky Maybe move documentation in `webpack` docs and just put link with small description in readme?

michael-ciniawsky:
@BurkovBA Yep, it is to 'verbose' in the sense that it's more an SSR guide in general with a specific solution unrelated to `css-loader` in many ways (`webpack` config, `webpack-node-externals` etc). This projects README focuses on the loader itself (`options` etc) without too much integration docs for different setups. Not to criticise the guide in general, but it would need to be a link to a guide like this then, to aviod bloating the README to much

michael-ciniawsky:
or a blog post in the sense of 'guide'

BurkovBA:
@michael-ciniawsky , @evilebottnawi 

Updated the PR.

I dropped everything non-CSS related. Also made it concise and less like a guide and more like a documentation.

evilebottnawi:
@BurkovBA sorry for delay, can you rebase your PR and use `mini-css-extract-plugin` plugin in example

evilebottnawi:
After investigate i think it is real out of scope of `css-loader`, better create issue in `webpack` docs repo and writing page about SSR, because it is require additional setup not only for `css-loader`. Anyway thanks for the PR.

PR #680, opened 2018-02-21
fix(getLocalIdent): generate `localIdent` consistently across platforms

kerhong:
Made output of getLocalIdent platform independent

No

**Summary**
Made getLocalIdent normalize path separators to produce same output on all platforms.

No

**Other information**
This solution was taken from https://github.com/webpack/loader-utils/blob/master/lib/interpolateName.js#L58

### DIFF
--- a/lib/getLocalIdent.js
+++ b/lib/getLocalIdent.js
@@ -8,7 +8,7 @@ var path = require("path");
 module.exports = function getLocalIdent(loaderContext, localIdentName, localName, options) {
 	if(!options.context)
 		options.context = loaderContext.options && typeof loaderContext.options.context === "string" ? loaderContext.options.context : loaderContext.context;
-	var request = path.relative(options.context, loaderContext.resourcePath);
+	var request = path.relative(options.context, loaderContext.resourcePath).replace(/\\/g, "/").replace(/\.\.(\/)?/g, "_$1");
 	options.content = options.hashPrefix + request + "+" + localName;
 	localIdentName = localIdentName.replace(/\[local\]/gi, localName);
 	var hash = loaderUtils.interpolateName(loaderContext, localIdentName, options);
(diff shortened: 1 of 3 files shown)


kerhong:
I don't particularly like the changes I did to helpers.js in order to create this test. Perhaps you could suggest a better way to do the same test?

lib/getLocalIdent.js
```diff
@@ -8,7 +8,7 @@ var path = require("path");
 module.exports = function getLocalIdent(loaderContext, localIdentName, localName, options) {
 	if(!options.context)
 		options.context = loaderContext.options && typeof loaderContext.options.context === "string" ? loaderContext.options.context : loaderContext.context;
-	var request = path.relative(options.context, loaderContext.resourcePath);
+	var request = path.relative(options.context, loaderContext.resourcePath).replace(/\\/g, "/").replace(/\.\.(\/)?/g, "_$1");
```
michael-ciniawsky:
```
path.posix.relative
```

?  But not sure...

michael-ciniawsky:
@kerhong This is likely to be breaking setups and we intent to remove CSS Modules from `css-loader` in the next major bump, so the CSSM implementation is in maintenance mode with 'critical' bugfixes only.

jackfranklin:
@michael-ciniawsky is there a rough plan with how CSS Modules will be supported in webpack longer term?

I ask as we use them with css-loader heavily, and would like to start preparing for moving off that and onto something else :)

michael-ciniawsky:
Sadly no current ETA and I always got distract by other things aswell :). I will try to streamline this effort and will post a RFC link here as soon as one is available

michael-ciniawsky:
I'm declining this PR. As mentioned a while ago we can not be sure if this breaks existing setups or not and we want to remove CSS Modules as quickly as possible anyways since the current implementation is problematic in general

PR #681, opened 2018-02-22
fix(getLocalIdent): add `rootContext` support (`webpack >= v4.0.0`)

yavorsky:
Bugfix.

No, because this project doesn't have e2e tests to test such case.

**Summary**
This PR fixes the behavior of `getLocalIdent`, which cause bugs with webpack 4.
[According to the chanelog](https://github.com/webpack/webpack/releases/tag/v4.0.0-beta.0), the `rootContext ` was added.  Loaders may use it to make stuff relative to the application root.

Also, `loaderContext.options` was removed, so the `options.context` is always value from `loaderContext.context`, which is actually filename without path.

So for example, if we have `a/a.css` and `b/a.css` both with `.c {}` content, the hash will be generated according to the `a.css` and `c` values (the same for these 2 independent files), which will cause class name conflicts.

Previous webpack versions don't include `loaderContext.rootContext` and as a result, this change is fully compatible with previous webpack versions and won't break anything.

No

**Other information**

### DIFF
--- a/lib/getLocalIdent.js
+++ b/lib/getLocalIdent.js
@@ -6,7 +6,7 @@ var loaderUtils = require("loader-utils");
 var path = require("path");
 
 module.exports = function getLocalIdent(loaderContext, localIdentName, localName, options) {
-	if(!options.context)
+	if(!options.context) {
 		if (loaderContext.rootContext) {
 			options.context = loaderContext.rootContext;
 		} else if (loaderContext.options && typeof loaderContext.options.context === "string") {
@@ -14,6 +14,7 @@ module.exports = function getLocalIdent(loaderContext, localIdentName, localName
 		} else {
 			options.context = loaderContext.context;
 		}
+	}
 	var request = path.relative(options.context, loaderContext.resourcePath);
 	options.content = options.hashPrefix + request + "+" + localName;
 	localIdentName = localIdentName.replace(/\[local\]/gi, localName);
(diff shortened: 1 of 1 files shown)


lib/getLocalIdent.js
```diff
@@ -6,8 +6,15 @@ var loaderUtils = require("loader-utils");
 var path = require("path");
 
 module.exports = function getLocalIdent(loaderContext, localIdentName, localName, options) {
-	if(!options.context)
-		options.context = loaderContext.options && typeof loaderContext.options.context === "string" ? loaderContext.options.context : loaderContext.context;
+	if(!options.context) {
```
evilebottnawi: Can we solve this in one line?

yavorsky: I could refactor it to nested ternary expressions:
`options.context = options.context || (loaderContext.rootContext ? loaderContext.rootContext : loaderContext.options && typeof loaderContext.options.context === "string" ? loaderContext.options.context : loaderContext.context)`

evilebottnawi: @yavorsky it is better in future to maintenance, just add or remove new options/context and etc

yavorsky: But we could just remove condition case.
Nested ternary expressions are not allowed by current eslint config.

evilebottnawi: let's wait @michael-ciniawsky 

Good work! Thanks!

michael-ciniawsky:
Released in [`v0.28.10`](https://github.com/webpack-contrib/css-loader/releases/tag/v0.28.10) 🎉

yavorsky:
Cool, thanks!

PR #684, opened 2018-02-26
feat: add `resultHook` (`options.resultHook`)

kossnocorp:
When resultHook is passed with the query, css-loader will call it with the loader context and the result. It will allow extending css-loader functionality without using a fork. For instance, it will be possible to generate TypeScript typing files on the fly.

This PR introduces a feature.

Yes.

Yes.

**Summary**
I want to add TypeScript support to [decss](https://github.com/kossnocorp/decss). To do so I have to generate type definition files. I can't do it via additional loader as I need an access to raw exports object.

Nope.

**Other information**
¯\\_(ツ)_/¯

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ It's useful when you, for instance, need to post process the CSS as a string.
 |**[`camelCase`](#camelcase)**|`{Boolean\|String}`|`false`|Export Classnames in CamelCase|
 |**[`importLoaders`](#importloaders)**|`{Number}`|`0`|Number of loaders applied before CSS loader|
 |**`localIdentName`**|`{String}`|`[hash:base64]`|Configure the generated ident|
+|**[`resultHook`](#resulthook)**|`{Function}`|`undefined`|CSS process result callback|
 
 ### `root`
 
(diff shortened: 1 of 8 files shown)


kossnocorp:
Well, it's sad to hear. Any alternatives to achieve that?

evilebottnawi:
Why don't parse output here https://github.com/webpack-contrib/css-loader/blob/master/lib/loader.js#L135 ?

kossnocorp:
@evilebottnawi are you suggest to eval css-loader's output? I can see how this could help but I'm afraid there are will be too many corner cases to make it stable.

evilebottnawi:
@kossnocorp look here in theory we can use `meta` to pass result. But in future we remove `css-modules` from loader in favor `postcss-plugin` or `css-modules-loader`. You can help us with this :+1:

kossnocorp:
@evilebottnawi having `meta` would be great. If I properly understand webpack allows to pass additional data along with source? I'm not familiar with this feature, can you please direct me to the docs or examples of usage?

Also, @michael-ciniawsky WDYT?

evilebottnawi:
@kossnocorp no docs :smile: Yep, for pass own data from loader to other loader

kossnocorp:
@michael-ciniawsky what do you think, won't it be out of scope?

PR #715, opened 2018-05-14
Allow query.getCssModuleMapping to get at css module classes at buildtime

jackfranklin:

Feature.

No, because I'd first like to get feedback on this idea - but I am happy to write tests :)

Again, no because I'd like to get a sense on if this PR will get merged. If you're open to merging it then yes I'll update the PR with docs.

**Summary**

The motivation here is that alongside importing our CSS modules in JS, we'd also like to write a `css-modules.json` file that we'll use for some tooling. Therefore, we wanted to maintain the behaviour of being able to write `import styles from '..'`, and get an object back, but also we wanted to hook into the build to get access at build time to the map of classes.

This PR introduces `getCssModuleMapping` which is called with the source file, and the map of classes, so that as a consumer I can hook into this and do anything I want with these files.

No.

**Other information**

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -108,9 +108,14 @@ module.exports = function(content, map) {
 				return "\" + escape(require(" + loaderUtils.stringifyRequest(this, urlRequest) + ")) + \"";
 			}.bind(this));
 		}
-		
+
+
 		var exportJs = compileExports(result, importItemMatcher.bind(this), camelCaseKeys);
 		if (exportJs) {
+			if (typeof query.getCssModuleMapping === "function") {
+				var fileName = loaderUtils.getRemainingRequest(this).split("!").pop()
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
Sry but I have to decline this as we intend to decouple/remove CSS Modules as soon as possible from the `css-loader` and provide them as a separate (postcss) plugin instead.

Your use case is on the list of features to be supported by that plugin e.g

```js
const modules = require('postcss-modules')

modules({
    json (exports) {
       // Handle it yourself...
    },
    // and/or
    emit: true || '[filename].json'
})
```

There is sadly no concrete ETA atm for this as we are rethinking the whole CSS Modules story in general detail, but also have a *lot*  of user we need to support with the current status quo which is difficult to organize :)

Issue #716, opened 2018-05-17
NPM audit - Critical Command Injection


MichaelFBA:
Because of a deep dependancy `macaddress` it is advised not to use this plugin.

```
Critical      │ Command Injection                                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ macaddress                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ No patch available                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ css-loader [dev]                                             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ css-loader > cssnano > postcss-filter-plugins > uniqid >     │
│               │ macaddress                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/654
```

TomS-:
Noticed this, is there an alternative until this is fixed?

MichaelFBA:
Looks like it might be getting fixed downstream.

evilebottnawi:
@MichaelFBA need help with release `cssnano@4`

evilebottnawi:
https://github.com/ben-eb/cssnano/milestone/6

mickael-kerjean:
Same here but this issue is a symptom of a much deeper problem: why would a dependency of a dependency of a dependency of a dependency of a plugin that manage CSS in a build system would need to know about my mac address in the first place?

The randomid version of the Leftpad syndrom? Lack of third party dependency audit? It doesn't feel right.

TomS-:
@mickael-kerjean I was thinking exactly that, why does it need to know my mac-address?

mickael-kerjean:
@JamesCodes : The question isn't what it does but why did this land here in the first place. You can generate a unique id in 1 line of code:
``` javascript
function uniqid(){ return (Math.random()*Math.pow(10,18)).toString(16) }
```

JamesCodes:
@michael-ciniawsky completely agree, it's not needed. Looks like a fix may be close: https://github.com/postcss/postcss-filter-plugins/pull/18

evilebottnawi:
Guys i am contributor and maintainer `cssnano`, feel free to send PR, don't have time right, go ahead :+1:

JamesCodes:
[...]

PR #718, opened 2018-05-20
Update postcss-filter-plugins version 3.0.1

BraedenNorris:
This was to mitigate the critical vulnerability with older versions of the postcss-filter-plugins that would allow command injection.

I updated version for a plugin that was causing a vulnerability.
No.
No.
**Summary**
Will stop vulnerability and errors from showing up when a user installs css-loader. #716 discusses this vulnerability.

No
**Other information**


evilebottnawi:
@BraedenNorris why you need update `lock` file? Please update own lock file. Thanks!

PR #740, opened 2018-07-04
chore(package): drop Node.js 4 support

ai:

refactoring

no relevant

no relevant

**Summary**

Nodde.js 4 has no security updates and was dropped a few Webpack versions ago. It is safe to drop this support.

We need it for cssnano 4 update https://github.com/webpack-contrib/css-loader/pull/739

But I do not really force this changes. If there is a reason why we can’t drop Node.js 4 support, we can found another way.

yes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: node_js
 branches:
   only:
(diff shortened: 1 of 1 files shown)


ai:
Close for  #742

PR #742, opened 2018-07-04
1.0

evilebottnawi:

refactor

already exists

yes

**Summary**

Roadmap:
1. Merge this PR and release `1.0.0` version.
2. Drop css modules in favor https://github.com/css-modules/postcss-modules
3. Refactor code (get some usefull stuff from https://github.com/webpack-contrib/css-loader/pull/639)
4. Release `2.0.0` version.

yes

**Other information**

Let's do it


ai:
Sure. What I need to look?

evilebottnawi:
@ai need find way how use `postcss-loader` with `postcss-modules` plugin (our pipeline https://github.com/webpack-contrib/css-loader/blob/master/lib/processCss.js#L189), maybe some values will be unavailable inside `postcss-loader`. 

In short: need add example `postcss-modules` configuration with `postcss-loader` in README.

ai:
👍 will send you a PR in next few days

.travis.yml
```diff
@@ -7,20 +7,20 @@ matrix:
   fast_finish: true
   include:
     - os: linux
-      node_js: "7"
-      env: WEBPACK_VERSION="2.2.0" JOB_PART=lint
+      node_js: "10"
```
shellscape: can we ditch travis and get on circle CI?

lib/loader.js
```diff
@@ -6,18 +6,14 @@ var loaderUtils = require("loader-utils");
...
 
 
 module.exports = function(content, map) {
-	if(this.cacheable) this.cacheable();
 	var callback = this.async();
 	var query = loaderUtils.getOptions(this) || {};
-	var root = query.root;
-	var moduleMode = query.modules || query.module;
-	var camelCaseKeys = query.camelCase || query.camelcase;
+	var moduleMode = query.modules;
```
shellscape: because this is a refactor, shouldn't we use `const` and `let` instead?

package.json
```diff
@@ -36,6 +34,9 @@
     "should": "^11.1.2",
     "standard-version": "^4.0.0"
   },
+  "peerDependencies": {
+    "webpack": "^3.0.0 || ^4.0.0"
```
shellscape: new major versions of webpack-contrib and webpack org modules are only supporting webpack@4 moving forward. since this is a new major version, we should drop webpack@3 support.

ev1stensberg: Agree with Andrew, it's harder to keep compat when things move forward too

evilebottnawi: @shellscape :+1:

evilebottnawi:
@shellscape i think we can update webpack-default after merge this PR, because problem with `browserslist` are ASAP

.travis.yml
evilebottnawi: @shellscape let's do this in other PR

lib/loader.js
evilebottnawi: @shellscape also let's do it other PR, when we ship webpack-default in this repo

.travis.yml
shellscape: OK

lib/loader.js
shellscape: OK

sokra:
There are few more breaking changes possible, but that should be considered carefully:

## Dropping `~` requests, urls -> requests

Currently it behaves like this:

`url("./something")` -> `require("./something")`
`url("something")` -> `require("./something")`
`url("~something")` -> `require("something")`

We can consider dropping this logic:

`url("./something")` -> `require("./something")`
`url("something")` -> `require("something")`

The logic was there for best compatiblity with existing stylesheets, but this may not be thing for everybody.

As workaround one can do this:

``` js
module.rules: [
  { test: /\.css$/,
    use: [MCEP.loader, "css-loader"],
    resolve: { modules: [".", "node_modules"] }
  }
]
```

which will change the logic to this:

`url("./something")` -> `require("./something")`
`url("something")` -> `require("something")` but looks in the current directory too.

For backward-compat `~` could be supported but deprecated. Or a separate loader can do the transform for backward-compat.

## Dropping `importLoaders`

Currently when using `@import` (and it's done by the css-loader, which is not the case when using sass or postcss-import) X loaders are copied to the inner require. X = `importLoaders`.

This is important to not apply the style-loader to inner requests, because that doesn't work.

When dropping `importLoaders` you have to manually make sure that `style-loader` is not appied to inner requests, i. e. with these rules:

``` js
module.rules: [
 { test: /\.css$, rules: [
    { issuer: { not: /\.css$/ }, use: "style-loader" },
    { use: "css-loader" }
 ]}
]
```

Similar rule for MCEP.

## ESM exports

When using "locals" (modules) the locals could be exported as ESM exports. This would make the result a ESM. For backward-compat expose it as named exports and all exports as object via default export.

package.json
```diff
@@ -36,6 +34,9 @@
     "should": "^11.1.2",
     "standard-version": "^4.0.0"
   },
+  "peerDependencies": {
+    "webpack": "^4.0.0"
```
ooflorent: Not sure about this. The codebase does not `require('webpack')`. I understand that you added this to avoid runtime errors when webpack's version is not the expected one but I don't think this is the correct solution.

evilebottnawi: @ooflorent we use this for all webpack-contrib repos, better create issue in `webpack-default` about this.

evilebottnawi: @ooflorent Also it is prevent problem with using webpack api between version, if somebody want use `css-loader` on `webpack@3`, they can use old version

shellscape: yeah it's a standard for webpack-defaults. that peerDep on webpack has been in defaults for a very long time. along with the `Requirements` section in the defaults README, this serves as a warning to keep people on supported versions of webpack. new major versions of loaders/plugins/utils are webpack@4+ and Node 6+ only.

Issue #652, opened 2018-01-02
Weird source map paths when used together with sass-loader


jwldnr:
**Do you want to request a *feature* or report a *bug*?**
Bug report

**What is the current behavior?**
Wrong source map/resource paths using `css-loader` together with `sass-loader` as shown below:

**Examples of source map paths using the plugin**
The actual `app.scss` path is `./src/styles/app.scss`

![ex1](https://user-images.githubusercontent.com/11960421/34483589-cdf4a6da-efbf-11e7-8178-de675034230c.png)

However the source map path for the file becomes `./src/styles/src/styles/app.scss` and the absolute resource path: `C:/Users/m51n/source/repos/source-map-path/src/styles/src/styles/app.scss` (does not exist)

![ex2](https://user-images.githubusercontent.com/11960421/34483600-e66bd0e4-efbf-11e7-8cb3-af945560996e.png)

**If the current behavior is a bug, please provide the steps to reproduce.**
I've created a sample project [here](https://github.com/jwldnr/extract-text-issue)-- using the source map example from [the `sass-loader` docs](https://github.com/webpack-contrib/sass-loader#source-maps) and code from the `extract-text-webpack-plugin` docs [here](https://github.com/webpack-contrib/extract-text-webpack-plugin#url-resolving)

With the sample project above I used `devtool: "inline-source-map"`.

**What is the expected behavior?**
Expected source map path returned by the plugin in this case would be `./src/styles/app.scss` and absolute resource path as it exist on disk: `C:/Users/m51n/source/repos/source-map-path/src/styles/app.scss`

**Related issues for reference**
https://github.com/webpack-contrib/sass-loader/issues/484

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**

webpack v3.10.0
node v9.2.0
os: windows 10

dcatoday:
Hi I am seeing an issue using ~paths.  The path is returned without any deliminating /'s in the url.

Songworks:
Came upon the same today with having duplicate-ish paths in sourcemaps.
Here's the relevant loaders in my config (no ExtractTextPlugin and devtool as *source-map*):
```javascript
  [
    {
      loader: 'style-loader',
    },
    {
      loader: 'css-loader',
      options: {
        sourceMap: true,
      },
    },
    {
      loader: 'sass-loader',
      options: {
        sourceMap: true,
      },
    },
  ];
```

[...]

PR #753, opened 2018-08-03
fix: source map `sources` and `file` paths

npetruzzelli:
half bugfix / half feature

Not yet, but they will be similar to existing tests. Waiting for feedback/comments.

Yes

**Summary**
The `sources` and `file` entries in sourcemaps are URLs relative to the file contains the source map (`.css` or `.map`) per version 3 of the source map specification.

For many loaders concerned with stylesheets, they currently try to treat them as file system paths, which has been a significant factor in errors that appear in these paths, especially when paths are being resolved without first checking if they are already absolute. These file system paths often end being relative to something other than the file containing the map.

"Under the hood" this loader uses PostCSS, which actually doesn't know about webpack's current request and remaining request. It only cares about file system paths. Part of the fix involves setting both the `from` (unmodified) and the `to` argument to being `loader.resourcePath`.

Using file system paths in maps is possible, but it is best left to the plugin responsible for writing the source map to the css file or the map file. The `style-loader` will have some additional considerations that are unique to it.

Once `css-loader` and `sass-loader` are fixed, `postcss-loader` and others will eventually need PRs as well.

Because the fix is a breaking change, but people may want to use it immediately (such as testing it with other loaders needing similar fixes), it has been implemented as a new option.

No, but the intent is a future major version could make the opt-in fix the default behavior.

**Other information**
This contributes to the fixes for
-   https://github.com/webpack-contrib/css-loader/issues/652
-   https://github.com/webpack-contrib/sass-loader/issues/529


npetruzzelli:
See also: https://github.com/webpack-contrib/sass-loader/pull/602

npetruzzelli:
Looks like the appveyor build is broken.

```
Build started
git clone -q https://github.com/webpack-contrib/css-loader.git C:\projects\css-loader
git fetch -q origin +refs/pull/753/merge:
git checkout -qf FETCH_HEAD

The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.
```

npetruzzelli:
@evilebottnawi - I've removed the option part of it. I still need to update the test suite and create a test repo. It may take me some time to get to it. I'll comment again once it is all in place. You may see commits trickle in in the mean time.

lib/processCss.js
```diff
@@ -179,16 +179,8 @@ module.exports = function processCss(inputSource, inputMap, options, callback) {
...
-	if (options.query.legacySourceMaps !== false) {
-		// we need a prefix to avoid path rewriting of PostCSS
-		postCssFrom = "/css-loader!" + options.from;
-	} else {
-		postCssFrom = options.from;
-	}
-
 	pipeline.process(inputSource, {
-		from: postCssFrom,
+		from: options.from,
```
evilebottnawi: Break maps in many cases

michael-ciniawsky:
First thx @npetruzzelli for the effort, but before we continue implementing we need to discuss the source map story as a whole. I had mutliple issue(s) about source maps since I'm maintaining these loaders and there were fixes provided that solved issues for some folks while adding new problems for others, it either is something which needs to be configurated for some cases or we just not getting it right atm... 😛

lib/processCss.js
npetruzzelli: Are you referring to the existing test suite? It is expected that it broke. I held off on updating the test suite as the comments in this review would determine how it would need to be updated.

The existing test suite is still using assumptions about source maps prior to the fix in this PR. I currently have the feedback I need and I intend to update the test suite prior to setting up the test repo.

If it isn't the test suite, is there a specific scenario you can describe that I can look into?

npetruzzelli:
@michael-ciniawsky - I am happy to discuss it. If I wanted a solution to work just for me I could publish a fork under a new namespace. My goal is to fix it for everyone. I see source maps as a valuable tool and these loaders will be the only option available to many who don't have the time/budget to learn how to make build systems from the ground up.

Can you share with me what the past issues were? The fix I'm envisioning sticks as close as possible to the source maps v3 spec. That said, if it breaks the tool for many users, then it isn't helping. If the use cases that need separate configuration can be clearly defined, I think it is a problem we can solve.

I am simply starting with css-loader and sass-loader. I can't speak for the JS source maps (yet), but I intend to touch more loaders and plugins related to the stylesheet source maps. The problem is not something you can point at one loader or plugin and say it is the source of the problem, at least with what I have found so far with css, sass, and postcss loaders. I'm not interested in pointing fingers, so I apologize if my choice of words seem harsh. I just want to help fix it!

michael-ciniawsky:
> If I wanted a solution to work just for me I could publish a fork under a new namespace. My goal is to fix it for everyone.

I cause my comment sounds like if I implied your currently trying to do that, that's not the case

> Can you share with me what the past issues were?

Sure, I will try to look some of them up, it was always about paths being incorrect (relative in the wrong way etc) in a back and fourth manner

- mainly from Sass (`sass-loader`)
- in conjunction with `resolve-url-loader`

The fix you propose in this PR already looks familiar to me (setting `options.from`/`options.to` to `loader.resourcePath`) and I think I used/tested it already and it worked to me knowledge, but it may break for some cases again, I haven't verify that part. I'm currently doing this in [`postcss-loader`](https://github.com/postcss/postcss-loader/blob/master/lib/index.js#L108) and not to long ago there was PR which removed setting it for `options.to`, setting only the `options.from` value, but this broke `resolve-url-loader` :D

npetruzzelli:
> > If I wanted a solution to work just for me I could publish a fork under a new namespace. My goal is to fix it for everyone.
> 
> I cause my comment sounds like if I implied your currently trying to do that, that's not the case

To clear things up, I didn't mean to make that implication! My PR effectively comes in and says that a past contributor got something wrong. My goal isn't to put that person down, I am here to help solve a challenge.

> mainly from Sass (sass-loader) in conjunction with resolve-url-loader

I'm having difficulty figuring out what `resolve-url-loader` is supposed to do. Is it trying to transform instances of `url()` from being a path relative to the CSS file to a path that is relative document root? The `resolve-url-loader` also appears to be using file system paths (relative and absolute) instead of URLs. I'd have to understand the module developer's intent better before speculating further. In all cases, `sass-loader` would need to have its own part taken care of first. Early loaders need to be fixed before later loaders.

I hope to more time into the pull requests for `css-loader` and `sass-loader` either this weekend or early next week. This week has been really busy both in and out of work.


> The fix you propose in this PR already looks familiar to me (setting options.from/options.to to loader.resourcePath) and I think I used/tested it already and it worked to me knowledge, but it may break for some cases again, I haven't verify that part

I took a quick look at postcss-loader and the related part of PostCSS's code. Even if this kind of trick works right now, there is no guarantee that it will work in the future as it is an undocumented implementation detail. PostCSS's own documentation is what says to set both values to the same path.  At least for when I've used the Autoprefixer PostCSS plugin, I've always received expected source maps when providing both `to` and `from` values in this manner.

ocoka:
**Although !** maybe I don't understand that PR correctly and my problem related to behavior of *stylus-loader* ... just tested in my repo, and my problem still there

npetruzzelli:
Appveyor is failing in a way that I don't believe is related to the pull request.

TravisCI is failing because `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`. It looks like npm stalled for Node.js version 6 and it never had the opportunity to run the suite.

@evilebottnawi - I've updated the test suite. My next task is to setup a test repo. I plan to take another pass at the test suite after feedback. Do you have any specific criteria you would like to see in the test repo?

@michael-ciniawsky - Have I addressed the concerns for this repository? (not including updates other loaders or plugins, I intend to handle those within their respective repos). I'll be happy to continue the conversation!

lib/loader.js
```diff
@@ -33,10 +54,31 @@ module.exports = function(content, map) {
...
+	 * (suchs as map.file or map.sources).
+	 *
+	 * Changing the file extension counts as changing the location because it
+	 * changes the path.
+	 *
+	 * PostCSS's `from` and `to` arguments are only concerned with the file
+	 * system. They don't know about, care about, or understand the webpack
+	 * loader's current request or remaining request.
+	 */
+	processCssFrom = processFrom(this.resourcePath, map);
```
```
var from = getFrom(this.resourcePath, map)
var to = from
```

lib/loader.js
```diff
@@ -33,10 +54,31 @@ module.exports = function(content, map) {
...
+	processCssTo = processCssFrom;
+
 	processCss(content, map, {
 		mode: moduleMode ? "local" : "global",
-		from: loaderUtils.getRemainingRequest(this).split("!").pop(),
-		to: loaderUtils.getCurrentRequest(this).split("!").pop(),
+		from: processCssFrom,
```
```js
{
  ...,
  // TODO use shorthand in next major
  from: from,
  to: to
}
```

lib/loader.js
```diff
@@ -2,18 +2,39 @@
...
+/**
+ * If the file was "renamed" (for the purposes of source maps), then honor it,
+ * otherwise just use the resourcePath.
+ * @param {String} resourcePath - The absolute file system path for the sass file.
+ * @param {Object|null} map - An existing source map, if any.
+ * @return {String} - The effective path to use for the `from` argument.
+ */
+function processFrom(resourcePath, map) {
+	var effectiveResourcePath;
+	if (map && map.file && typeof map.file === 'string' && path.dirname(map.file) === '.') {
```
michael-ciniawsky: `path.dirname(map.file) === '.'`  checks for a relative path here ?  I'm not sure I'm following.. 😛 . In case my understanding is correct here, there is also `path.isAbsolute(filepath)` available

lib/loader.js
```diff
@@ -2,18 +2,39 @@
...
+		// Something else has already changed the file name or extension, so
+		// honor it for the purpose of creating the next source map.
+		effectiveResourcePath = path.join(path.dirname(resourcePath), map.file);
```
michael-ciniawsky: Please provide an example for `effectiveResourcePath` here :)

lib/loader.js
```diff
@@ -2,18 +2,39 @@
...
+	} else {
+		effectiveResourcePath = resourcePath;
+	}
```
michael-ciniawsky: nitpick 🐦 `\n` after the `else` block

lib/loader.js
```diff
@@ -112,13 +154,6 @@ module.exports = function(content, map) {
 		if(sourceMap && result.map) {
 			// add a SourceMap
 			map = result.map;
-			if(map.sources) {
```
michael-ciniawsky: Can we safely remove this ? `sourceRoot` should be empty (not sure if it can be non existent) and the `file` and `sources` paths are normalized here (L116+). Is this a bug ? :) What happens on windows platforms ?

Please elaborate why you removed it

npetruzzelli:
@michael-ciniawsky - I'll try to have complete responses for you in the next day or so. Sorry for the delay!

michael-ciniawsky:
Take your time no pressure at all, but we need to be quite pendantic here 😛

lib/loader.js
npetruzzelli: I've done some thinking on it and here I may be trying to be too clever. What I originally was thinking of shouldn't be something that css-loader is responsible for. I can provide more details on what I was thinking if you'd like.

I'll try cleaning this up in my next commit.

```javascript
		from: this.resourcePath,
		to: this.resourcePath,
```

will be enough.

lib/loader.js
npetruzzelli: Short answer: **no, the removal of this code must remain.**

---

`sourceRoot` is something that the developer responsible for creating / maintaining the build system should be managing. We should not be modifying it.

Whatever is consuming the map will prepend the value of sourceRoot to each entry in sources to create a URL. If it is not an absolute URL, then it treats the path as a relative URL that is relative to the **map** not the CSS file, though this may be one and the same if the map is inlined into the CSS file.

I need to double check this next part, but I think I remember seeing that some uses of `sourceRoot` will modify the resulting paths in the `sources` array, meaning it has an impact on how future map generators and consumers handle including existing maps into newly generated maps.

---

The normalizing you are referring to attempted to do multiple things. The first is account for use of Webpack's currentRequest and remainingRequest in the `from` and `to` fields. That is going away and is no longer relevant.

The remaining normalizing attempts to treat `sources` as file system paths, which is wrong. Under normal use, they are URLs being consumed by a web browser to request source files from a web server.

The `path` module is one of node's core modules for working with file system paths. not URL paths.
---

In a completely closed build system, it may be possible to use file system paths, but Webpack and its loaders are not closed, they are open, as are the processors that the loaders and plugins consume. Since URLs is what the spec defines, we must provide URLs for other tools to work as expected.

source map writers (style-loader, mini-css-extract-plugin, etc.) may provide an option to output the final maps as absolute file system paths because they will be capable of resolving relative URLs to relative file system paths, and ultimately, absolute file system paths. in this scenario, the writer would want to remove the sourceRoot property or set it to be the `file:` URL protocol/URI scheme.

The are multiple reasons for this, but each have to do with the assumption that the file writer is the last thing to modify the resource as it will take into account the output destination. Remember, changing the location from the source directory to the build directory is also a transformation, so the writer should be creating an updated map of its own. We see this working correctly with [`gulp.dest()`]( https://github.com/gulpjs/gulp/blob/49b5aca613b33c5b626ae68c03a385f25c142f55/docs/API.md#gulpdestpath-options ) and [`gulpSourcemaps.write()`]( https://github.com/gulp-sourcemaps/gulp-sourcemaps/tree/492d937ffa0d5aa62a0d5b99ceb7deb2f472c7ad#write-options ).

---

For whatever reason, the Webpack community has tried to treat the URLs as absolute file system paths at every step. They also treat `sourceRoot` differently, when it shouldn't be treated at all. Some try to make paths relative to the Current Working Directory under the assumption that it is also the host/document root directory. Many of these loaders and plugins are passing mangled maps to tools like node-sass, postcss, and babel. They in turn, output bad maps or may ignore the mangled map entirely, and it continues to get worse with each node module touching the source map like this.

**Why has it kind of worked up until this point?**

-   sometimes the current working directory actually is the host root (but this is never guaranteed, and we shouldn't be using file system paths anyway)
-   when you embed source content in the map, the browser uses that instead of trying to fetch the source file. This why a number of guide may recommend this, because "it just works"
    -   Developers are left to make reasonably educated guesses on the actual location of the source file based on what ever value makes it into the sources array. If this value is an unmangled absolute file system path, this gives the appearance of it working correctly when it is not.
    -   Certain developers tools and plugins in the browser may be able to actually access the file system (by default or through permissions and configuration), but since Webpack is not a closed system, we can never assume this.

I am absolutely able to support this through experience, results, and referencing the specification and other places I've made inquiries on for this topic.

This is part of why I can say with absolute confidence that this is not just a single loader's problem. I'm starting with css-loader and sass-loader, but I see a lot of work to be done before this gets to where it needs to be in the webpack community.

Apologies for the rant! I hope it didn't come off as defensive. I'm just a bit nervous because I have seen past open issues effectively say "there is nothing wrong with this repository's code, go open an issue in the other module's repo". Since the problem spans multiple code bases, I will need support from contributors in each of those repositories. So part of this is trying to make the argument for it as strong and as clear as possible.

I will be happy to explain/converse more if I didn't clear it up. I'm unfortunately better at speaking code to computers than speaking English to people, so by all means, ask me as much as you need to! I'll do my best to be thorough and constructive.

michael-ciniawsky: > sourceRoot is something that the developer responsible for creating / maintaining the build system should be managing. We should not be modifying it.

'`webpack`' [`SourceMapDevToolPlugin`](https://github.com/webpack/webpack/blob/master/lib/SourceMapDevToolPlugin.js) modifies it in the end, to my (limited) knowledge a `map`'s `sourceRoot` returned fro a loader should be either `''` or `undefined`

> The normalizing you are referring to attempted to do multiple things. The first is account for use of Webpack's currentRequest and remainingRequest in the from and to fields. That is going away and is no longer relevant.

👍 

> Under normal use, they are URLs being consumed by a web browser to request source files from a web server.

> The path module is one of node's core modules for working with file system paths. not URL paths.

> In a completely closed build system, it may be possible to use file system paths, but Webpack and its loaders are not closed, they are open, as are the processors that the loaders and plugins consume. Since URLs is what the spec defines, we must provide URLs for other tools to work as expected.

> source map writers (style-loader, mini-css-extract-plugin, etc.) may provide an option to output the final maps as absolute file system paths because they will be capable of resolving relative URLs to relative file system paths, and ultimately, absolute file system paths. in this scenario, the writer would want to remove the sourceRoot property or set it to be the file: URL protocol/URI scheme.

`webpack` uses it's own (devtools) protocol `webpack://` for source maps, which are URL's but `sources` ideally should be absolute paths for the `SourceMapDevToolPlugin`to work correctly. That's what I was at least told when asking about it the last time :)

For the `style-loader` it may be a different story since we can inline the map in the `<style></style>` tags as an annontation comment, kind of 'bypassing'  `webpack` internals

michael-ciniawsky: > For whatever reason, the Webpack community has tried to treat the URLs as absolute file system paths at every step. They also treat sourceRoot differently, when it shouldn't be treated at all. Some try to make paths relative to the Current Working Directory under the assumption that it is also the host/document root directory. Many of these loaders and plugins are passing mangled maps to tools like node-sass, postcss, and babel. They in turn, output bad maps or may ignore the mangled map entirely, and it continues to get worse with each node module touching the source map like this.


Again likely because of `SourceMapDevToolPlugin` and because nobody _really_ knows how to handle it properly by default nor for anykind of variation xyz :)

> This is part of why I can say with absolute confidence that this is not just a single loader's problem. I'm starting with css-loader and sass-loader, but I see a lot of work to be done before this gets to where it needs to be in the webpack community.

True, it's an general problem within this area atm

npetruzzelli: > 'webpack' (SourceMapDevToolPlugin) modifies it in the end, to my (limited in particular) knowledge a map's sourceRoot returned from a loader should be either '' or undefined

The `SourceMapDevToolPlugin` does modify it. I'll need to look at it more closely again. There is a lot going on so it will take some time for me to figure it out. This plugin is one of the "writers" I plan to make contributions to in the future (if needed), but for the last loader/plugin to be fixed, it must receive a valid Source Map from everything that has come before it.

---

> webpack uses it's own (devtools) protocol webpack:// for source maps, which are URL's, but sources ideally should be absolute paths for the SourceMapDevToolPluginto work correctly. That's what I was at least told when asking about it the last time :)
> 
> For the style-loader it may be a different story since we can inline the map in the <style></style> tags as an annotation comment (\n//# sourceMappingURL=[url]) directly, kind of 'bypassing' webpack internals
> 
> So from a plugin/loader authors perpective we need to provide the correct map values the SourceMapDevToolPlugin needs rather then the 'final' result

I am aware of the devtools option. My understanding so far is that it is a wrapper around `SourceMapDevToolPlugin` for less complex use cases.

I have tried to get more information about `webpack://` through StackOverflow per the instructions presented in when opening an issue in Webpacks repo. I was met largely with silence and downvotes, which ultimately lead to the auto deletion of the question.

From my own investigation, the scheme only appears so that it can be a default value for two configuration options: 
https://github.com/webpack/webpack/blob/4f4a2bae33fa1f8b1954aa0bbfd5b9e3bef95cf8/lib/SourceMapDevToolPlugin.js#L70-L74

**`webpack:` is not a valid registered URI scheme ( <https://www.iana.org/assignments/uri-schemes/uri-schemes.xml> ).** No web browser would know what to do with it. Beyond this, there is no publically information available to me about why this is used by webpack's core code.

`SourceMapDevToolPlugin` requiring absolute file system paths as input is wrong, though I haven't seen any documentation stating this as a requirement. Even if `SourceMapDevToolPlugin` knows what to do with them, other modules, which loaders and plugins often consume, such as PostCSS, Sass, Babel, and NYC/Istanbul, are not expecting absolute paths, nor is there any requirement from the Source Maps specification for them to do so. This can contribute to further mangling of the source maps.

`SourceMapDevToolPlugin` outputting absolute file system paths is not something I would object to, so long as it is a configurable option that is disabled by default. This kind of behavior should be opt in. Since it is effectively the last piece of software to perform a transformation, it is ok for the output to be whatever we want it to be. In doing so, the developer using our plugin or loader takes responsibility for it being different from the specification.

The `style-loader` has unique considerations which I will need a lot more time to learn about. This is thanks in part to it serving blobs. Support for hot module replacement will also add to the complexity. `webpack-dev-server` (and the middleware it consumes) will probably need to be tackled at the same time as the `style-loader`. Since URLs are normally relative URLs to the file containing the map (even inlined maps), serving the blob is effectively another transformation since you are changing the file's depth/path relative to the host root directory.

---

> Again likely because of SourceMapDevToolPlugin and because nobody really knows how to handle it properly by default nor for anykind of variation xyz :)

I have seen tools in the Gulp community handle it correctly, though there are some that don't. There is no reason why Webpack can't ultimately be the same. I think I have a reasonably good idea on how to handle it, my efforts here are an attempt to start down that path.

---

I will be happy to consider any of my current knowledge wrong if I can be pointed to specifications or documentation. What I have seen in other issues in and out of this repo boiled down to "I think that's how it works" or "this person told me this was how it works, but didn't cite any sources." Unless that person wrote the spec, that isn't enough for me to back down. So far, I haven't found any sources proving me wrong, just opinions and guesses. So I had to get the ball rolling with a pull request of my own.

npetruzzelli: @michael-ciniawsky - I just wanted to check in since it has been about a week. No rush in replying! I just wanted to make sure I had touched on everything you were looking for a response on. 

For reference, @evilebottnawi mentioned in an issue in the sass-loader repo that source maps are not expected to be tackled until the next major revision of the loader. Is that also true for the css-loader?

Should I [resolve conversation](https://help.github.com/articles/commenting-on-a-pull-request/#resolving-conversations) to collapse this series of comments and replies and return to the main thread of the PR for discussion?

npetruzzelli:
@evilebottnawi @michael-ciniawsky - Thank you both for taking the time to consider this PR. I don't know if a month is a long time or not for an open source PR to not receive an update or comment. Much like the rest of us, I'm sure there is alot going on for you both to balance between work, family, friends, and open source. It really is harder than it looks! I've resolved open threads. Please let me know if you'd like to discuss this PR (and/or the sass-loader one) more in the future.

Best Regards.

evilebottnawi:
@npetruzzelli work was not stopped, right now i am working on https://github.com/webpack-contrib/css-loader/pull/760, after release new version we will review your PR and give feedback, sorry for delay, a lot of works, but many developers :disappointed:

npetruzzelli:
Got it. I'll keep an eye out for your next response, though I understand it may be some time. Thank you for your response!

npetruzzelli:
I'm in the process of merging in the most recent updates to:
https://github.com/webpack-contrib/css-loader/tree/master
into
https://github.com/npetruzzelli-forks/css-loader/tree/source-map-uris

I think I was able to carry over the bulk of my changes as a part of resolving the merge conflicts, but changes to unit testing means I have to update the test suite again. Once I update the test suite, I will be able to confirm whether or not this PR is still working as intended. I'm used to working with Mocha/Chai/Sinon. It may take a bit for me to wrap my head around Jest.

evilebottnawi:
@npetruzzelli Great! If you have problems with jest, send a PR with tests, we can write this late :+1:

npetruzzelli:
@evilebottnawi 

I am still working on this, but found a couple things that are confusing me: 

```javascript
// lib/loader.js

   /* eslint-disable no-param-reassign */
   if (sourceMap) {
     if (map) {
       if (typeof map === 'string') {
         map = JSON.stringify(map);
       }
```

### 1. Why is `no-param-reassign` being disabled?

Isn't it generally considered a bad practice to reassign named arguments? Especially with V8 because it prevents code optimization by the engine/compiler? Does this only happen if the arguments object itself is accessed within the same function?

### 2. If the map is already a string, why are we stringifying it?

Valid types for PostCSS's `prev` property on the map option includes: String, Object, Boolean, and Function. It isn't stated, but I imagine `undefined` is valid, but I'm not sure about `null`.

See: <https://github.com/postcss/postcss/blob/08cbd147c840e7ea54ea4a036acabd0673129e7a/docs/source-maps.md>

evilebottnawi:
> Why is no-param-reassign being disabled?

We can refactor this logic. Also this place for buggy loaders. 

>  If the map is already a string, why are we stringifying it?

Some loader pass string, some objects. Also just compatibility.

lib/loader.js
```diff
@@ -100,15 +97,30 @@ module.exports = function loader(content, map) {
...
   postcss(plugins)
     .process(content, {
-      // we need a prefix to avoid path rewriting of PostCSS
-      from: `/css-loader!${getRemainingRequest(this)
-        .split('!')
-        .pop()}`,
-      to: getCurrentRequest(this)
-        .split('!')
-        .pop(),
+      from: resourcePath,
```
evilebottnawi: It is dangerous using this, also it is invalid use `resourcePath`, because it is not resource path, it is request, maybe you can create problem repository?

npetruzzelli:
The following is a diagram illustrating where in PostCSS that code relating to source maps may be found. I may be using it as a reference later on in this code review.

<https://github.com/postcss/postcss/tree/046ab11f1be0eb3dd62bb3ea331134119168af11>

```
PostCSS (lib/PostCSS.es6)
├── parse (same as PostCSS -> processor -> lazy-result)
└─┬ processor
  └─┬ lazy-result
    ├─┬ map-generator
    │ └── mozilla (is the source-map module)
    ├─┬ parse
    │ └─┬ input
    │   └─┬ previous-map
    │     └── mozilla (is the source-map module)
    └── result (this.map is an instance of `SourceMapGenerator` class from the `source-map` library, representing changes to the Result#root instance.)
```

lib/loader.js
npetruzzelli: It will take me a while to completely respond to this. If I am lucky, I can have something ready some time this week.

I need to understand more about the differences between when to use the request and when to use a file system path.

---

It is difficult to learn the details of the request syntax, because it does not exist on the website (<https://webpack.js.org/concepts/>) nor the documentation used to generate it (<https://github.com/webpack/webpack.js.org>). The closest we come is documentation on context: <https://webpack.js.org/api/loaders/#the-loader-context>.

There are some pages in the old wiki, but they don't go into a lot of detail and users looking for documentation might not expect to look in the deprecated docs repository wiki because they appear to have the same information at first glance and would expect the syntax to be documented on the site. In reality, the wiki contains information not present in the in the more up to date documentation: <https://github.com/webpack/docs/wiki/loaders>, <https://github.com/webpack/docs/wiki/using-loaders>.

---

The next post will probably be a big one. 

Before I make it, I want to see if what I'm concerned about is still an issue or if I'm worried about nothing. My next step is to create the example repository that you requested as it will help illustrate what has been fixed and what still needs work. I can't rely only on the test suite just yet.

npetruzzelli:
## Valid `prev` values

### SUMMARY

-   PostCSS accepts a number of formats, but represents the maps internally as a stringified object.
-   Internally, Webpack expects loaders to expose source maps as plain objects
    -   This seems to conflict with the documentation for `this.callback` which accepts any type that is acceptable to Mozilla's **source-map** module.
-   Niether PostCSS's nor Webpack's requirements and implementations are a hinderance to each other.

_The details here are informational only. No action or response is being requested at this time. The details contain a TODO item to be looked at in the future._

<details>

### DETAILS

I'm considering making the following reccomendation:

-  For simplicity, the previous source map should be either stringified _OR_ left unmodified when being supplied to PostCSS.

but I don't have enough information to make it worth someone else's effort to review it. The current state of the code would have no impact on the issues this PR attempts to solve.

The current code stringifies something that is already a string, but otherwise leaves the previous source map unmodified. This may or may not contribute to the issues in a subtle way. Currently, there are other items that need to be looked at before this.

#### PostCSS

> **`prev` string, object, boolean or function:** source map content from a previous processing step (for example, Sass compilation). PostCSS will try to read the previous source map automatically (based on comments within the source CSS), but you can use this option to identify it manually. If desired, you can omit the previous map with prev: false.
> 
> _Source: [PostCSS and Source Maps](https://github.com/postcss/postcss/blob/38095826f973fccff8641d188c80e0f58f661472/docs/source-maps.md#options) documentation for options._

Publically, PostCSS accepts a number of possibilities:

-   The boolean literal `false` (`true` is not valid). This ignores previous source maps
-   A the source map object as a `string`
-   A `function` that is passed the **options.map.from** option as an argument. The returned result is expected to be a file system path that points to an existing map file. It uses Node's own **fs** module, so other file systems, such as **memory-fs**, are not supported.
-   An `object` that is an instance of `SourceMapConsumer` from Mozilla's **source-map** module.
-   An `object` that is an instance of `SourceMapGenerator` from Mozilla's **source-map** module.
-   A plain `object` that represents a source map.

Internally, PostCSS represents sourcemaps as strings. Providing a string will avoid some condition checking and processing ([source code](https://github.com/postcss/postcss/blob/6b426af97f0db3d8d2e4d2615b2cb79b08b4d91a/lib/previous-map.es6#L96-L133)).

#### Webpack

> The first loader is passed one argument: the content of the resource file. The compiler expects a result from the last loader. The result should be a String or a Buffer (which is converted to a string), representing the JavaScript source code of the module. An optional SourceMap result (as JSON object) may also be passed.
> 
> _Source: [Loader API](https://github.com/webpack/webpack.js.org/blob/d5f9b865ccc92fc731197312b7fb7d6051110e9b/src/content/api/loaders.md) introduction._

> 1. The first argument must be an `Error` or `null`
> 2. The second argument a `string` or a [`Buffer`](https://nodejs.org/api/buffer.html).
> 3. Optional: The third argument must be a source map that is parsable by [this module](https://github.com/mozilla/source-map).
> 4. Optional: The fourth option, ignored by webpack, can be anything (e.g. some meta data).
> 
> _Source: [Loader API: this.callback](https://github.com/webpack/webpack.js.org/blob/d5f9b865ccc92fc731197312b7fb7d6051110e9b/src/content/api/loaders.md#thiscallback)_

</details>

npetruzzelli:
## PostCSS's `options.map.from` And `options.map.to` Options

### SUMMARY

1.  For the most dependable results, an absolute file system path should be passed as the value to this option.
2.  Since the css-loader is not responsible for writing the css files or source maps, the `to` value should be identical to the `from` value

_The details here are informational. No review of the above reccomendations are needed at this time as they may change once the reproducible test repo has been created and reviewed. The purpose of this post is to capture information and links that may be relevant to future decision making._

<details>

### DETAILS

If **options.map.from** is not provided, PostCSS will attempt to determine the the file's location using the **file** property on the source map. This can lead to unexpected results as since it will silently fall back to the current working directory, which isn't necessarily going to be the correct location of the source file. (source: [`input Constructor`](https://github.com/postcss/postcss/blob/84bbc4979f2c6c07cdfbbcb732419abc49b961b1/lib/input.es6#L43-L75), [`input.mapResolve`](https://github.com/postcss/postcss/blob/84bbc4979f2c6c07cdfbbcb732419abc49b961b1/lib/input.es6#L146-L151)).

If we provide **options.map.from** as a relative path, it is subject to a similarly unexpected result.

Both scenarios use Node.js's own `path.resolve` method is used to create an absolute file system path. It is this method that falls back to the current working directory to create an absolute file system path if the result of the function doesn't end up as one through the normal course of path resolution.

> **If after processing all given path segments an absolute path has not yet been generated, the current working directory is used.**
> 
> The resulting path is normalized and trailing slashes are removed unless the path is resolved to the root directory.
> 
> Zero-length path segments are ignored.
> 
> _Source: Node.js [path.resolve([...paths])](https://nodejs.org/api/path.html#path_path_resolve_paths)_

> To ensure that you generate an accurate source map, you **must** indicate the input and output CSS **file paths** — using the options `from` and `to`, respectively.
> 
> _Source: [PostCSS and Source Maps](https://github.com/postcss/postcss/blob/046ab11f1be0eb3dd62bb3ea331134119168af11/docs/source-maps.md)_


> 2.1. Set from and to processing options  
> To ensure that PostCSS generates source maps and displays better syntax errors, runners must specify the `from` and `to` options. **If your runner does not handle writing to disk (for example, a gulp transform), you should set both options to point to the same file:**
> 
> ```
> processor.process({ from: file.path, to: file.path })
> ```
> 
> _Source: [PostCSS Runner Guidelines](https://github.com/postcss/postcss/blob/829cb8511175c7bf7f019fad7189a021ae3de4c0/docs/guidelines/runner.md#21-set-from-and-to-processing-options)_

The webpack request syntax is not a valid argument for PostCSS's `from` and `to` options. Attempting to fix it and force it may work in some situations, but creates invalid source maps in others.

</details>

lib/loader.js
evilebottnawi: @npetruzzelli feel free to create minimum reproducible test repo, it is allow to me also identify the problem and how to fix it

evilebottnawi:
I know how `postcss` works (i am contributor of `stylelint`, `prettier`, `cssnano` and etc, which used `postcss`). Let's create repo with problem and i find solution how we should solve this problem

lib/loader.js
npetruzzelli: Hi @evilebottnawi, 

I've created the repository that illustrates the problem. I'll need some time to get some text and images together to support/describe it clearly instead of needing you to dig through multiple pull requests and issues across multiple repositories for a clear picture.

https://github.com/npetruzzelli-forks/webpack-loader-demos 

I plan to use this repository to demonstrate the eventual solution just by pointing to a different Webpack file, probably in a new branch.

Relevant information:
-   OS: **Windows 10 Pro 64-bit**
-   NODE: **v8.9.4**
-   NPM: **6.4.1** (see also: [npm-windows-upgrade](https://github.com/felixrieseberg/npm-windows-upgrade))
-   Windows Build Tools: **[5.0.0](https://github.com/felixrieseberg/windows-build-tools/tree/a1ff38f579f77a205269e9bb6e6f3de5cd0ea2fa)**

evilebottnawi: @npetruzzelli great, thanks, i will see tomorrow, very tired today

npetruzzelli: No rush! I know that feeling. Rest well.

npetruzzelli:
@evilebottnawi : It looks like the master branch has been updated and the file structure has changed again. I'll need time to handle merge conflicts. 

---

Here are some notes for the demonstration repository that I eventually plan to add to that repository's README:


## NPM Scripts:
-   `npm run build-all`: Run all builds sequentially
-   `npm run fullstack-build`: Build a combination of **Sass + PostCSS + CSS** loaders, no fixes applied, to `.tmp/dist/full-style-stack/`
-   `npm run fullstack-serve-dev`: Serve a combination of **Sass + PostCSS + CSS** loaders, no fixes applied
-   `npm run lint`: Run ESLint
-   `npm run postcss-build`: Build a combination of **PostCSS + CSS** loaders, no fixes applied, to `.tmp/dist/postcss-stack/`
-   `npm run postcss-serve-dev`: Serve a combination of **PostCSS + CSS** loaders, no fixes applied
-   `npm run sass-build`: Build a combination of **Sass + CSS** loaders, no fixes applied, to `.tmp/dist/sass-stack/`
-   `npm run sass-serve-dev`: Serve a combination of **Sass + CSS** loaders, no fixes applied
-   `npm run serve-src`: Serve files from the source directories. Intended to be used in conjunction with Webpack's **devServer.proxy** option.
    -   **TODO:** serve **node_modules** from this task as well.
-   `npm run start`: runs both **fullstack-serve-dev** and **serve-src** scripts.
-   `npm run styles-build`: Build a combination of **CSS** loaders, no fixes applied, to `.tmp/dist/css-stack/`
-   `npm run styles-serve-dev`: Serve a combination of **CSS** loaders, no fixes applied

## Expected / Actual

When running the server and inspecting the paragraph of red lorem ipsum text, and examining the `.my-lead-text` rule in the developer console, the source line is not expected, accurate, or predicatable. Without an accurate and predicatable value, the proxy server can not be reasonably configured to serve source code files or files from node_modules for deeper inspection.

_Note: the difference in port numbers are intentional._

-   **EXPECTED:**
    -   `../../../../../../src/_assets/stylesheets/styles.scss`, line 5 (Sass)
    -   `../../../../../../src/_assets/stylesheets/styles.css`, line 1 (CSS)
-   **CSS + PostCSS + Sass:** (actual)
    -   `http://localhost:9008/_assets/bundles/main/src/_assets/stylesheets/D:/Code/webpack-loader-demos/src/_assets/stylesheets/styles.scss`, line 5
-   **CSS + Sass:** (actual)
    -   `http://localhost:9006/_assets/bundles/main/src/_assets/stylesheets/styles.scss`, line 5
-   **CSS + PostCSS:** (actual)
    -   `http://localhost:9004/_assets/bundles/main/src/_assets/stylesheets/styles.css`, line 1
-   **CSS:** (actual)
    -   `http://localhost:9002/_assets/bundles/main/styles.css`, line 1

npetruzzelli:
Please note that the demonstration repo does not yet include webpack configurations that demonstrate a fix.

npetruzzelli:
For the sake of definition the `.tmp` ("temp" or "temporary") directory is a directory that will contain:
-   files that are may be safely deleted at any time
-   built files that will never be put under version control, this may include
    -   dist or "distribution" files: compiled bundles, scripts, styles, etc.
    -   coverage files: code coverage reports generated from software tests
    -   build related cache files, such as `.sass-cache`, if applicable

---

For the sake of definition the `src` (source) directory is a directory that will contain:
-   all files which may be considered an entry point (JavaScript or otherwise).
-   source code that needs to be processed or transformed
-   source code that this repository is responsible for maintaining

---

For the sake of definition the `static` directory (also known as the "public" directory) is directory that will contain:
-   files that this repository is responsible for maintaining, but **don't** need processing or transformation to be consumed. Examples of this includes a favicon, robots.txt, and humans.txt.
-   files that this repository **is not** responsible for maintaining, but still needs to consume. Examples of this includes third party libraries that can't be or won't be managed by software like bower, npm, or yarn
-   When bundled, the only action taken on these files should be copying some or all of them from one location to another.

This directory is often considered to be a secondary source for an HTTP server's document root directory in development environments.

---

For the sake of definition the `test` directory is directory that will contain:
-   test suite related files, including spec files
    -   some projects may have a convention defined that places unit test spec files in the same **src** directory as the code under test. In these cases the test directory may still contain other kinds of tests, including end-to-end/integration tests.

---

For the sake of definition the `tools` directory is directory that will contain:
-   configuration for development tools that don't need to be in the repository root directory
-   utilities for these development tools

npetruzzelli:
Updated by:

-   [Disambiguate the `[name]` replacement](https://github.com/npetruzzelli-forks/webpack-loader-demos/commit/f30a1f4518a0636e5a8ba0efec9d78fbe6b7f0d2)
-   [Source server updates](https://github.com/npetruzzelli-forks/webpack-loader-demos/pull/1)

The new commit for the demo repo's master branch is:
https://github.com/npetruzzelli-forks/webpack-loader-demos/tree/298e7e56ea093733ce39677d89a34c25858f6582

evilebottnawi:
@npetruzzelli thanks, we fix it in near future, many works right now, sorry for delay

npetruzzelli:
@evilebottnawi - I can try to set some time aside to rebase this branch and resolve any conflicts. I wasn't sure (based on the last comment in December) if you were working on a separate fix. I apologize for not seeking clarification.

Once I am caught up, I'll try to see if the demo repo can be updated/improved as well. Either way it shouldn't delay the rebase.

I'll try to tackle it at some point in the next few weeks. If I'm lucky, I may have time on Sunday to work on this.

npetruzzelli:
@evilebottnawi - I've begun working on updating this branch, but I've run into an error on commit: 

```
git -c diff.mnemonicprefix=false -c core.quotepath=false --no-optional-locks commit -q -F C:\Users\UserName\AppData\Local\Temp\42ob4uyp.xi1
husky > pre-commit (node v8.9.4)
Stashing changes... [started]
Stashing changes... [skipped]
→ No partially staged files found...
Running linters... [started]
Running tasks for *.js [started]


eslint --fix [started]
eslint --fix [completed]
git add [started]
git add [completed]
Running tasks for *.js [completed]
Running linters... [completed]


Warning: Setting commit-msg script in package.json > scripts will be deprecated
Please move it to husky.hooks in package.json, a .huskyrc file, or a husky.config.js file
Or run ./node_modules/.bin/husky-upgrade for automatic update

See https://github.com/typicode/husky for usage

husky > commit-msg (node v8.9.4)






Using environment variable syntax ($GIT_PARAMS) in -e |--edit is deprecated. Use '{-E|--env} GIT_PARAMS instead'
D:\Code\css-loader\node_modules\@commitlint\cli\lib\cli.js:109
	throw err;
	^

Error: Received $GIT_PARAMS as value for -e | --edit, but GIT_PARAMS is not available globally.
    at getEditValue (D:\Code\css-loader\node_modules\@commitlint\cli\lib\cli.js:242:10)
    at normalizeFlags (D:\Code\css-loader\node_modules\@commitlint\cli\lib\cli.js:212:15)
    at D:\Code\css-loader\node_modules\@commitlint\cli\lib\cli.js:116:11
    at new Promise (<anonymous>)
    at main (D:\Code\css-loader\node_modules\@commitlint\cli\lib\cli.js:113:9)
    at Object.<anonymous> (D:\Code\css-loader\node_modules\@commitlint\cli\lib\cli.js:105:1)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)










husky > commit-msg hook failed (add --no-verify to bypass)
Completed with errors, see above.
```

I'll start investigating the error, but I was wondering if you might know how to resolve it. It is preventing me from committing.

evilebottnawi:
@npetruzzelli remove node_modules and reinstall their

npetruzzelli:
@evilebottnawi - That was the first thing I tried. I tried it again just to be certain, but no luck.

I tried the following: 

-   Deleting everything except for the `.git` directory and then discarded changes.
-   `npm install`

It appears to be working now, though it looks like there is a new commit message syntax that I'll have to follow, that is easy enough to do. I'll have a look at the CONTRIBUTING markdown.

I suspect the issue may have been in the generated `dist` directory. I should be able to resume work on this at the next available opportunity.

evilebottnawi:
The main goal of this PR it testing source map on `windows`/`maos`/`linux` with/without `postcss-loader`, we can avoid `sass-loader` testing here (it is part of `sass-loader` tests)

evilebottnawi:
Looks we need do this manually and add tests step by step

lib/loader.js
evilebottnawi: Reproducible test repo is not very simple, i try around 30 minutes to understand structure. Also looks many things broken, js source maps also doesn't work in you example. Please simplify example.

evilebottnawi:
Done in https://github.com/webpack-contrib/css-loader/pull/901, will be released in near future, feel free to tests and if you faced with issue feel free to create new issue with reproducible test repo, thanks for PR

PR #562, opened 2017-06-23
feat: support imported classes in selectors (`postcss-icss`)

apexskier:
Still work in progress, opening for visibility and guidance. Currently works locally but due to dependency updates tests don't pass.

## What kind of change does this PR introduce?
Adds support for importing css modules in selectors. I guess it's a new feature, since this wasn't supported, but I'd consider it more of a bug fix, since the feature is part of the spec.

## Did you add tests for your changes?
- [ ] add tests
- [ ] fix broken tests

## Summary
This add support for importing css selectors from other locally scoped css files. This feature has been requested a few places and, in my opinion, allows css modules to use one of the core features that makes css great: being able to style components contextually.

- https://github.com/css-modules/css-modules/issues/102
- closes https://github.com/css-modules/icss/issues/12, closes https://github.com/webpack-contrib/css-loader/issues/561

A quick example of usage (using bare icss syntax):
```css
/* btn.css */
.btn { /* ... */ }

/* grouped-btn.css */
:import("./btn.css") {
    btn: btn;
}
.grouped-btn .btn {
    /* changing stuff, maybe removing a border radius on inner .btns */
}
```
Previously, this would have been compiled to something like:
```css
/* btn.css */
.__btn-style__btn { /* ... */ }

/* grouped-btn.css */
.__grouped-btn-style__grouped_btn .__grouped-btn-style__btn { /* ... */ }
```
Note that the `.btn` class has been scoped locally to the grouped button styles.

Now, this compiles to:
```css
/* btn.css */
.__btn-style__btn { /* ... */ }

/* grouped-btn.css */
.__grouped-btn-style__grouped_btn .__btn-style__btn { /* ... */ }
```

## Does this PR introduce a breaking change?
Possibly, this updates dependencies, and the base `icss-utils` repo has had a major version bump. I've got to investigate why tests are failing to know for sure. I'll update if so.

The main changes are based on https://github.com/css-modules/icss-utils/commit/19ffee9f4071aacabb3a49ca01783de0fad96b88

Also, thanks to @michael-ciniawsky for some context here - see https://github.com/webpack-contrib/css-loader/issues/561#issuecomment-310769225

PR #760, opened 2018-08-24
Next

evilebottnawi:
bugfix, feature, refactoring

Yes

Yep

**Summary**

Rewrite `css-loader`

Yes


evilebottnawi:
Looks node on `appveyor` generate other hashes when `linux`, very weird, maybe problem in https://github.com/css-modules/generic-names (we should update loader-utils to latest version)

evilebottnawi:
what the hell with CI :confused:

michael-ciniawsky:
Ideally we extract the **needed** `interpolateName` code from `loader-utils` and use it in `generic-names` directly or add a new method to one of the ICSS Utils and use that instead. The dependency on `loader-utils` is not needed

michael-ciniawsky:
Hashes are not platform consistent to my knowledge, but maybe it was fix in newer version of `loader-utils`

evilebottnawi:
@michael-ciniawsky wip on updating

joepie91:
Just gave this a quick testrun, and I ran into the following error:

```
ERROR in ./test.css
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
TypeError: Cannot read property 'foo' of undefined
    at Object.../3rdparty/css-loader-upstream/dist/cjs.js?!./test.css (/home/sven/projects/3rdparty/css-loader-upstream/dist/cjs.js??ref--5-1!/home/sven/projects/icss-loader-test/test.css:786:251)
    at __webpack_require__ (/home/sven/projects/3rdparty/css-loader-upstream/dist/cjs.js??ref--5-1!/home/sven/projects/icss-loader-test/test.css:684:30)
    at /home/sven/projects/3rdparty/css-loader-upstream/dist/cjs.js??ref--5-1!/home/sven/projects/icss-loader-test/test.css:751:93
    at Object.<anonymous> (/home/sven/projects/3rdparty/css-loader-upstream/dist/cjs.js??ref--5-1!/home/sven/projects/icss-loader-test/test.css:754:10)
    at Module._compile (/home/sven/projects/icss-loader-test/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at exec (/home/sven/projects/icss-loader-test/node_modules/mini-css-extract-plugin/dist/loader.js:55:10)
    at childCompiler.runAsChild (/home/sven/projects/icss-loader-test/node_modules/mini-css-extract-plugin/dist/loader.js:133:14)
    at compile (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compiler.js:296:11)
    at hooks.afterCompile.callAsync.err (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compiler.js:553:14)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:24:1)
    at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/Hook.js:35:21)
    at compilation.seal.err (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compiler.js:550:30)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/Hook.js:35:21)
    at hooks.optimizeAssets.callAsync.err (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compilation.js:1294:35)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/Hook.js:35:21)
    at hooks.optimizeChunkAssets.callAsync.err (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compilation.js:1285:32)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/Hook.js:35:21)
    at hooks.additionalAssets.callAsync.err (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compilation.js:1280:36)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/Hook.js:35:21)
    at hooks.optimizeTree.callAsync.err (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compilation.js:1276:32)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/Hook.js:35:21)
    at Compilation.seal (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compilation.js:1213:27)
    at hooks.make.callAsync.err (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compiler.js:547:17)
    at _err0 (eval at create (/home/sven/projects/icss-loader-test/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:11:1)
    at _addModuleChain (/home/sven/projects/icss-loader-test/node_modules/webpack/lib/Compilation.js:1064:12)
 @ ./index.jsx 3:10-31
```

Full testcase can be found at https://git.cryto.net/joepie91/icss-loader-test. This compiled 'correctly' with `css-loader` 1.0.0 except for it producing the wrong class names due to #562, so presumably the CSS itself is correct.

I tested it against `next` by cloning the `css-loader` repository, switching to the `next` branch, `npm install`ing it, and `yarn link`ing it into my test repository. I didn't use `npm link` or install-from-git due to a litany of outstanding npm bugs. I've verified that the error really does occur in `next` and not in 1.0.0.

joepie91:
On closer inspection, it also seems to not support CSS Modules? Or at least the `composes` syntax doesn't seem to do anything, and ends up as-is in my build product (when removing the `:import` block to avoid the compilation error).

There doesn't seem to be any option to enable support for this either, and the option for specifying a class name ident format has also disappeared in this release.

evilebottnawi:
@joepie91 old css modules doesn't work with nex css modules (ICSS) in some cases, you can find example with css modules in README

evilebottnawi:
@joepie91 also you have invalid code you try to import class, but no export `:export{foo: foo}`

joepie91:
Hmm, the [ICSS spec](https://github.com/css-modules/icss#export) suggests that you should only specify exports for exported *literals*, not for class names?

> An `exportedValue` does not need to be quoted, it is already treated as a __literal string__.

(Emphasis mine.)

How does that combine with having to explicitly export classes? Given that those are presumably not literal strings.

evilebottnawi:
@joepie91 to avoid big size of `locals` you should export only what need, `css-loader@1` works same

michael-ciniawsky:
I will review it in-depth today, it overall looks very good I quickly went over it yesterday :). The bigger remaining things currently are 

- Move `@import/url()` handling to `css-modules/postcss-icss-import` && `css-modules/postcss-icss-url` aswell, so only message and token handling remains within the loader (add the plugins as dependencies)
- Usability story for CSS Modules via `postcss-loader` (via e.g a plugin) (TBD)
- https://github.com/webpack-contrib/css-loader/pull/760#issuecomment-416315794 (`composes` Bug (?)) (I will try it out myself later)

evilebottnawi:
@michael-ciniawsky I already answered all these questions :disappointed: Sometimes it seems to me that you just ignore my message.

> Move @import/url() handling to css-modules/postcss-icss-import && css-modules/postcss-icss-url aswell, so only message and token handling remains within the loader (add the plugins as dependencies)

- `postcss-icss-import/postcss-icss-url` is not compatibility with `ICSS` spec.
- We use other logic in some places, because we should `require` `url/import`.
- Also it is allow to fast fix `url/import` bugs.

> Usability story for CSS Modules via postcss-loader

Example already in import, creating one plugin is out of scope `css-loader`, if somebody want this, go ahead and send a PR to README

> #760 (comment) (composes Bug (?)) (I will try it out myself later)

Invalid usage `:import/:export`

michael-ciniawsky:
> I already answered all these questions 😞 Sometimes it seems to me that you just ignore my message.

No that's not the case, bI don't agree with some of them and will not sign-off the PR as-is if isn't beeing discussed how to handle those. If we still disagree (or can't find an agreement) that's fine but then someone else needs to additionally review the PR and approve it. 

> postcss-icss-import/postcss-icss-url is not compatibility with ICSS spec.

They both should be and what's outdated there compared to the plugins included in this PR simply needs to be upstreamed 

> We use other logic in some places, because we should require url/import

There are still loader specific things which need to be handled within the loader itself, but those should be mainly related to stringification of messages and tokens, where the `import/url` plugins should **only** provide the messages and tokens, so it's possible to use them elsewhere (e.g ) (decoupling)

> Also it is allow to fast fix url/import bugs

Since we have access to all repos we can always fix bugs 'with ease'  from now on :)

> Example already in import, creating one plugin is out of scope css-loader, if somebody want this, go ahead and send a PR to README

It is not directly part of this PR as the whole point here is to remove them from the loader, there also doesn't need to be a plugin asap (but a 'ROADMAP'  to a potental solution). We need to be sure that writing one is generally possible and works **beforehand**. The `postcss-loader` example for CSS Modules is definitely a usability downgrade compared to the current implementation `{ modules: true }` and pushes all kind of errors folks will have with that to the `postcss-loader`. I'm not willing to deal with that for some time simply because we didn't take the time to resolve it properly for no good reason. We don't have any time pressure at all here...

Yes, I'm willing to write this plugin myself (it was not meant to push the workload on you or the like)

> Invalid usage :import/:export

> On closer inspection, it also seems to not support CSS Modules? **Or at least the composes syntax doesn't seem to do anything**, and ends up as-is in my build product (when removing the :import block to avoid the compilation error).

I need to try it out myself, @joepie91 could you please elaborate further and clearify what exactly didn't work for you when using `composes` ? Is there a resolver/loader for `composes` ?


We could release `alpha/beta` versions of `css-loader >= v2.0.0`  at first and split remaining work while being able to iterate freely on the matter

evilebottnawi:
@michael-ciniawsky 

> There are still loader specific things which need to be handled within the loader itself, but those should be mainly related to stringification of messages and tokens, where the import/url plugins should only provide the messages and tokens, so it's possible to use them elsewhere (e.g ) (decoupling)

Why? What is place?

> They both should be and what's outdated there compared to the plugins included in this PR simply needs to be upstreamed

Go ahead and implement/update their and we replace on plugins. As soon as they are ready we will replace them with external plugins.

> It is not directly part of this PR as the whole point here is to remove them from the loader, there also doesn't need to be a plugin asap (but a 'ROADMAP' to a potental solution). We need to be sure that writing one is generally possible and works beforehand. The postcss-loader example for CSS Modules is definitely a usability downgrade compared to the current implementation { modules: true } and pushes all kind of errors folks will have with that to the postcss-loader. I'm not willing to deal with that for some time simply because we didn't take the time to resolve it properly for no good reason. We don't have any time pressure at all here...

> Yes, I'm willing to write this plugin myself (it was not meant to push the workload on you or the like)

Go ahead.

I don't see critical problems. Also I would like to say that the previous releases were devilishly abandoned and very often with a waste of time to talk. And very often because of the chatter and not the writing of the code. If you have a **practical solution** (**worked**) please show where we can use them.  Everything else can be fixed and updated in the next versions. While we do not try the solution, we can not say anything about its success. It is possible to spend time indefinitely on conversations, I prefer to write code, and even if I make a mistake somewhere, I just correct it and go ahead.

michael-ciniawsky:
> Why? What is place?

What do you mean here 🙃 ? Other places to use them ? Native CSS Module support in webpack ? Other bundlers like rollup && parcel ? When talking from an ICSS/CSS Modules point of view

### ICSS (Plugins)

#### 1.1

```
:import {}
:export {}
```

#### 1.2 
```
{
   type: 'import'
   ...,
   import (...args) { return `...` } 
}

{
   type: 'export'
   ...,
   export (...args) { return `...` } 
}
```

---- CSS Loader Specifics (or any other tools) start here ---

#### 2. Messages/Tokens Handling (`css-loader`)

```js
...

// ICSS Plugins
const icss = {
  url: require('postcss-icss-url')(options.url)
  import: require('postcss-icss-import')(options.import)
}

postcss([ icss.url, icss.import ])...

...

// Pseudo
// via ICSS Tokens
const tokens = ICSSUtils.extract(result.css)

tokens.forEach((token) => ...)

// via result.messages
result.messages.forEach((msg) => ...)
```

> Go ahead and implement/update their and we replace on plugins. As soon as they are ready we will replace them with external plugins.

Ok 


> Go ahead.

Ok

> I don't see critical problems. Also I would like to say that the previous releases were devilishly abandoned and very often with a waste of time to talk.

🙄 You keep repeating that, did you for e.g review them even once to provide some (initial) feedback? No? So did I solely contribute to that fact? I finally closed them after some time to avoid any extra effort after **you** started your own **without** wanting to pick up/build on what was already there (which is ok). What's the point and argument to blame me here? What's does fighting and complaining over that lost effort add here? Our source of disagreement is about last mile efforts for this PR. If you're (understandably) tired and aren't willing to address them let me 'take over' again and PR against this branch, so I'm handling the bikesheding

michael-ciniawsky:
> It looks like you did not even look at my code.

Of course did I look at the code :), you extract the tokens and transform them into message `{Object}`'s  `{ type: '', ...} right ? 1.1 && 1.2 are both ways to achieve the same result and I thought we already agreed that we will follow the ICSS 'Spec' (using tokens) here for now? I don't propose to revert that to use message (exclusively)

> We should send only tokens, not function Why? Because it is in ICSS spec also other plugins send only tokens, you purpose will be necessary to rewrite all existing postcss-icss-* plugins (waste of time again and will be abandoned again). Also we don't have resources on this global rewriting only because it seems to you that the functions are better (I repeat that this violates the ICSS spec).

Yes, the (postcss) plugins use ICSS tokens and the CSS Loader gets the CSS and extracts them, then converts them to message `{Objects}` concats them with existing messages and stringifies them. The `{Function}` was your idea and request since it seemed to be needed to provide (loader) context ? It's also possible to to this within the loader and don't use a `{Function}` on the message `{Objects}` themselves. My point here is that the plugin just return the ICSS tokens and can be moved to their own repo respectively (`postcss-icss-url` && `postcss-icss-import`) and any further loader specfic handling is done by the loader. It shouldn't be a big deal to do that refactor and again this makes the ICSS plugins for `@import/url()` usable by other tools

> We already do this, except that we have our own implementation due reasons above (incorrect currently implementation and some limitations).

Did I every say we don't ? Please provide more detailed info about what is incorrect and what are these limitations and why the can't be solved (with the current code of the PR) while we still _could_ extract the import/url plugins and partially refactor that part a bit to do so. Note that this doesn't mean we must do it (now)

 > I just can not understand you. 

I also can not understand you atm, as all I 'requested' is to polish and iterate on your PR and to discuss potentially remaining TODO's and eventual areas for improvement as per review e.g better usability for CSS Modules when used with `postcss-loader`. I'm aware this is not directly part of this PR, but it's still coupled to it since the PR removes CSS Modules. We need to test and verify the new way to use CSS Modules and ideally account for potential improvements we can make as the next steps. 'Here use the `postcss-icss-*` in some way with `postcss-loader` is not an equal replacement in comparison and can be done in a better way.

I fully realized the support of ICSS, covered all the tests (100%, all tests will be moved, noone deleted, exclude removed options), simplified many things as much as possible. I tested and tested many different approaches in finding a better solution, but you just repeat the same abstract theoretical solution what is not worked (or have big limitation), reject my solution just because you want to use the function and messages, ignore ICSS spec and all my time spent investigating on this. I do not want to be rude. But it does not look professional.

You misunderstand, I didn't reject it at all nor do I just want it my way. E.g I, from my point of view made the compromise by accepting the `{String}` based tokens approach to not escalate anything. Nobody is devaluing your work here, 

> But it does not look professional.

To that tendency of yours to sometimes make bolder claims about other peoples thoughts and intentions and insinuate unprofessionality (ad-hominem) when faced potential disgreement, I will just say that you sometimes should more self reflect on that. I never made any personal attacks, the whole shifting the discussion towards ad-hominem arguments is coming from your side atm. If I did personally attack you with anything in this thread then please list where and what (and why) 

> So I'm waiting review on code and not bikesheding.

Ok, this was badly formulated and agreed, but still the details are important here even if it sometimes gets annoying and exhausting to some point. It's not only exhausting and annoying for you btw :)

> Also I would like to note that sometimes you can just ask why exactly i implement this, rather than offer your theoretical pseudo code (this very clogs the thread and it is hard to read).

I thought about this and I agree that this can be exhausting and misunderstood and should refrain from doing so more. I will try to adjust my behaviour here in the future

michael-ciniawsky:
Yes and at least revamp the docs about how to use them a bit (I can do that). Should we release it as an `alpha/beta` version and as `css-loader@next` for a few days?

michael-ciniawsky:
I still need to review it, but time flyed by I hadn't had a chance yet :)

michael-ciniawsky:
Yeah, we would need to switch the default branch to `next` and add a note to the `README` that we want feedback on `yxz` and that the latest stable is `link/to/v1` which s arguably cumbersome. Feedback and adoption of `pkg@next` release was also always very low in my experience, I'm not really convinced by this pattern in general anymore... kk then let's just do a 'normal' release

evilebottnawi:
@michael-ciniawsky `next` will be used very few people because most users use `*-cli` (`vue-cli`, `creact-facebook-react-app` and etc) and they update packages only on stable. After this https://github.com/webpack-contrib/css-loader/pull/760#issuecomment-416903176 i will update code and ping your again

evilebottnawi:
All in master

PR #763, opened 2018-08-28
feat(loader): add `exclude` option (`options.exclude`)

daydayuphuan:
When I set `modules` to `true`, I want to ignore certain files, but I need to use these loaders again. I have no way to find a solution.

For example, my project uses `monaco-editor`, `monaco-editor` uses the import method in the component to load `css`, but it does not use the method like `react -> className` to set the class.

so I have to change it myself

add `exclude` to `options`

**webpack.config**
```
{
    test: /\.css$/,
    use: [
        {
            loader: 'style-loader'
        },
        {
            loader: 'css-loader',
            options: {
                modules: true,
                importLoaders: 1,
                localIdentName: '[name]_[local]',
                camelCase: true,
                exclude: [
                    path.resolve(__dirname, 'node_modules/monaco-editor')
                ]
            }
        },
        {
            loader: 'postcss-loader',
            options: {
                ident: 'postcss',
                plugins: function () {
                    return [
                        postcssImport,
                        postcssVars,
                        autoprefixer({
                            browsers: ['last 3 versions', 'Safari >= 8', 'iOS >= 8']
                        })
                    ];
                }
            }
        }
    ]
}
```

**Hope to adopt, thx**

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -8,14 +8,17 @@ var getImportPrefix = require("./getImportPrefix");
 var compileExports = require("./compile-exports");
 
 
-module.exports = function(content, map) {
+module.exports = function (content, map) {
 	var callback = this.async();
 	var query = loaderUtils.getOptions(this) || {};
 	var moduleMode = query.modules;
 	var camelCaseKeys = query.camelCase;
 	var sourceMap = query.sourceMap || false;
+	var exclude = query.exclude;
+	var from = loaderUtils.getRemainingRequest(this).split("!").pop();
+	var to = loaderUtils.getCurrentRequest(this).split("!").pop();
 
-	if(sourceMap) {
+	if (sourceMap) {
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
As a heads up we curretnly reworking the CSS Modules implementation as a whole and it will be removed from  the `css-loader` directly in `>= v2.0.0` (#760). Your feature request is a **must have** in the future but it's likely you will need to wait some time until it is implemented in an upcoming new plugin for CSS Modules we are working on. Let's discuss it for the current release line as proposed by your PR, but maybe you need to use your own fork in the meantime

evilebottnawi:
Problem still exists?

evilebottnawi:
Closing due to inactivity. Please test with latest version and feel free to reopen if still regressions. Thanks!

PR #777, opened 2018-09-28
docs(readme): add an example how to `@import` a node module

danny-andrews-snap:
The syntax for importing a node module is not easy to find (see https://github.com/webpack/webpack/issues/1789#issuecomment-277024235). Maybe this will help a few people out.

Dcoumentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,11 +115,16 @@ url(~module/image.png) => require('module/image.png')
 
 ### `import`
 
+To import styles from a node module path, prefix it with a `~`:
+
+```css
+@import '~module/styles.css';
+```
+
 To disable `@import` resolving by `css-loader` set the option to `false`
 
 ```css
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -119,6 +119,7 @@ To disable `@import` resolving by `css-loader` set the option to `false`
 
 ```css
 @import url('https://fonts.googleapis.com/css?family=Roboto');
+@import '~module/styles.css';
```
michael-ciniawsky: I don't think this belongs here since the example is for **disabling** `@import` resolving and why/when one would want to do that :)

danny-andrews-snap: Whoops! How does it look now?

danny-andrews-snap:
@michael-ciniawsky how does it look, now?

PR #791, opened 2018-11-17
Update postcss and mocha version

pierreneter:
Update postcss version for work with postcss plugins.
Update mocha because security warnings.


pierreneter:
Sorry, it already exists at: https://github.com/webpack-contrib/css-loader/tree/next

PR #797, opened 2018-11-26
docs: replace to-string-loader to raw-loader

JLHwung:

docs

yes

**Summary**
replace `to-string-loader` to `raw-loader` since it is also maintained in webpack-contrib.

**Other information**


evilebottnawi:
@JLHwung they are difference, `raw-loader` just load file, `to-string-loader` use `toString()` (https://github.com/gajus/to-string-loader/blob/master/src/to-string.js#L17), you can add new example

Issue #578, opened 2017-07-14
broken escaping


eeve:
**What is the current behavior?**
 The escape symbol `\` was lost
```css
.grid            { display: flex; flex-wrap: wrap; }
.grid.\-top      { align-items: flex-start; }
.grid.\-middle   { align-items: center; }
.grid.\-bottom   { align-items: flex-end; }
```
=>

```css
.grid            { display: flex; flex-wrap: wrap; }\n.grid.-top      { align-items: flex-start; }\n.grid.-middle   { align-items: center; }\n.grid.-bottom   { align-items: flex-end; }\n
```

**What is the expected behavior?**
Keep the escape symbol `\`

https://github.com/eeve/test.v1

evilebottnawi:
@eeve btw, why your use `\` in `selectors`

jschlieber:
I'm running into the exact same issue. In my case escape sequences are used in Predix-UI CSS Modules, e.g:

https://github.com/PredixDev/px-spacing-design/blob/db32341b44fee68cace50d5d236306f768bbed26/_trumps.spacing.scss#L175-L181

So that basically means I can't use those `u-m+` and `u-m++` classes.

Any ideas for a quick workaround?

evilebottnawi:
@jschlieber unfortunately no, try to found and fix problem today

jschlieber:
@evilebottnawi Any news on this issue?

evilebottnawi:
@jschlieber I was on vacation, in the near future I will take care of this, thanks for waiting

evilebottnawi:
@jschlieber `u-m+` and `u-m++` invalid css classes (https://jigsaw.w3.org/css-validator/validator), your should use `u-m\+` and `u-m\+\+`

evilebottnawi:
@jschlieber also `.\-top {}` and `.-top {}` are valid class selectors and mean the same thing.

evilebottnawi:
Bug inside `postcss-modules-local-by-default` and `postcss-modules-scope`, they remove `\` character. But i wonder use case. 

@jschlieber `.u-m\+` and etc selectors work good (https://github.com/webpack-contrib/css-loader/pull/584).

@TrySound what do your think about this?

jschlieber:
@evilebottnawi Ok, but what you're telling me is that i can't define my css selector as:
```css
.u-m++{
  ...
}
```
and I'm totally fine with that. But still, if classes are defined like this
```css
.u-m\00002b{
  ...
[...]

PR #894, opened 2019-02-15
fix: allow values in selectors to work correctly

jquense:
This PR contains a:

- [x] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

depends on: 
- https://github.com/css-modules/postcss-modules-values/pull/2#issuecomment-464125471
- https://github.com/css-modules/postcss-modules-scope/pull/4

This works to enable behavior that is specced by css-modules, but has been only partially implemented until now; **TL; DR;** selectors are a valid place for imported values to be interpolated.

### Breaking Changes

Maybe! Still thinking through that

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,15 +446,24 @@ for specific values in a css file that can be used in another.
 /* theme.css */
 @value primary: #BF4040;
 @value secondary: #1F4F7F;
+
+@value large: (min-width: 960px);
 ```
 
 Values are automatically exported and can be imported like so:
 
 ```css
-@value primary from './theme.css';
+@value primary, large from './theme.css';
 
-.text-primary {
+.header {
   color: primary;
+  padding: 0 10px;
+}
+
+@media large {
+  .header {
+    padding: 0 20px;
+  }
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -437,6 +437,46 @@ To import from multiple modules use multiple `composes:` rules.
 }
 ```
 
+##### `Values`
+
+`css-loader` also includes support for css-modules `@value`s which you can read more of here: https://github.com/css-modules/postcss-modules-values Values provide a way to declare import and exports
+for specific values in a css file that can be used in another.
+
+```css
+/* theme.css */
+@value primary: #BF4040;
+@value secondary: #1F4F7F;
+```
+
+Values are automatically exported and can be imported like so:
+
+```css
+@value primary from './theme.css';
+
+.text-primary {
+  color: primary;
+}
+```
+
+You also don't need to declare a an export as a value to reference in another file:
+
+```css
+/* button.css */
+
+.button {
+  color: red;
+}
+```
+
+```css
+@value button from './button.css';
+
+/* the 'button' class will be replaced with it's hashed value from 'button.css' */
+.toolbar > .button {
+  margin-left: 2rem;
+}
+```
+
 ### `localIdentName`
 
 Type: `String`
@@ -584,8 +624,8 @@ By default, the exported JSON keys mirror the class names. If you want to cameli
 
 |        Name        |    Type     | Description                                                                                                              |
 | :----------------: | :---------: | :----------------------------------------------------------------------------------------------------------------------- |
-|    **`false`**     | `{Boolean}` | Class names won't be camelized |
-|     **`true`**     | `{Boolean}` | Class names will be camelized, the original class name will not to be removed from the locals |
+|    **`false`**     | `{Boolean}` | Class names won't be camelized                                                                                           |
+|     **`true`**     | `{Boolean}` | Class names will be camelized, the original class name will not to be removed from the locals                            |
 |   **`'dashes'`**   | `{String}`  | Only dashes in class names will be camelized                                                                             |
 |    **`'only'`**    | `{String}`  | Introduced in `0.27.1`. Class names will be camelized, the original class name will be removed from the locals           |
 | **`'dashesOnly'`** | `{String}`  | Introduced in `0.27.1`. Dashes in class names will be camelized, the original class name will be removed from the locals |
--- a/src/plugins/postcss-icss-parser.js
+++ b/src/plugins/postcss-icss-parser.js
@@ -1,6 +1,6 @@
 import postcss from 'postcss';
 import valueParser from 'postcss-value-parser';
-import { extractICSS } from 'icss-utils';
+import { extractICSS, replaceValueSymbols } from 'icss-utils';
 import loaderUtils from 'loader-utils';
 
 const pluginName = 'postcss-icss-parser';
@@ -9,22 +9,22 @@ export default postcss.plugin(
   pluginName,
   () =>
     function process(css, result) {
-      const imports = {};
-      const icss = extractICSS(css);
-      const exports = icss.icssExports;
+      const importReplacements = Object.create(null);
+      const { icssImports, icssExports } = extractICSS(css);
 
-      Object.keys(icss.icssImports).forEach((key) => {
+      let index = 0;
+      Object.keys(icssImports).forEach((key) => {
         const url = loaderUtils.parseString(key);
 
-        Object.keys(icss.icssImports[key]).forEach((prop) => {
-          const index = Object.keys(imports).length;
+        Object.keys(icssImports[key]).forEach((prop) => {
+          index += 1;
 
-          imports[`$${prop}`] = index;
+          importReplacements[prop] = `___CSS_LOADER_IMPORT___${index}___`;
 
           result.messages.push({
             pluginName,
             type: 'icss-import',
-            item: { url, export: icss.icssImports[key][prop], index },
+            item: { url, export: icssImports[key][prop], index },
           });
 
           const alreadyIncluded = result.messages.find(
@@ -56,41 +56,45 @@ export default postcss.plugin(
           }
 
           const token = node.value;
-          const importIndex = imports[`$${token}`];
+          const replacement = importReplacements[token];
 
-          if (typeof importIndex === 'number') {
+          if (replacement != null) {
             // eslint-disable-next-line no-param-reassign
-            node.value = `___CSS_LOADER_IMPORT___${importIndex}___`;
+            node.value = replacement;
           }
         });
 
         return tokens.toString();
       }
 
-      // Replace tokens in declarations
-      css.walkDecls((decl) => {
-        // eslint-disable-next-line no-param-reassign
-        decl.value = replaceImportsInString(decl.value.toString());
-      });
-
-      // Replace tokens in at-rules
-      css.walkAtRules((atrule) => {
-        // Due reusing `ast` from `postcss-loader` some plugins may lack
-        // `params` property, we need to account for this possibility
-        if (atrule.params) {
+      // Replace tokens
+      css.walk((node) => {
+        if (node.type === 'decl') {
           // eslint-disable-next-line no-param-reassign
-          atrule.params = replaceImportsInString(atrule.params.toString());
+          node.value = replaceImportsInString(node.value.toString());
+        } else if (node.type === 'rule' && node.selector != null) {
+          node.selector = replaceValueSymbols(
+            node.selector,
+            importReplacements
+          );
+        } else if (node.type === 'atrule') {
+          // Due reusing `ast` from `postcss-loader` some plugins may lack
+          // `params` property, we need to account for this possibility
+          if (node.params) {
+            // eslint-disable-next-line no-param-reassign
+            node.params = replaceImportsInString(node.params.toString());
+          }
         }
       });
 
       // Replace tokens in export
-      Object.keys(exports).forEach((exportName) => {
+      Object.keys(icssExports).forEach((exportName) => {
         result.messages.push({
           pluginName,
           type: 'export',
           item: {
             key: exportName,
-            value: replaceImportsInString(exports[exportName]),
+            value: replaceImportsInString(icssExports[exportName]),
           },
         });
       });
--- a/src/plugins/postcss-icss-parser.js
+++ b/src/plugins/postcss-icss-parser.js
@@ -58,7 +58,7 @@ export default postcss.plugin(
           const token = node.value;
           const replacement = importReplacements[token];
 
-          if (replacement != null) {
+          if (replacement) {
             // eslint-disable-next-line no-param-reassign
             node.value = replacement;
           }
--- a/test/__snapshots__/modules-option.test.js.snap
+++ b/test/__snapshots__/modules-option.test.js.snap
@@ -5453,6 +5453,14 @@ Array [
   ],
   Array [
     3,
+    ".dSGVakTZ8YY-XNW64tg9F {
+  display: inline;
+}
+",
+    "",
+  ],
+  Array [
+    4,
     "._22ETk3ZhhvjVe-4dAULcsF {
   display: block;
 }
@@ -5469,6 +5477,9 @@ Array [
 .green > ._22ETk3ZhhvjVe-4dAULcsF {
   color: blue;
 }
+
+.uTNBxZ_JJrJOoGzCUbSnW > .dSGVakTZ8YY-XNW64tg9F {
+}
 ",
     "",
   ],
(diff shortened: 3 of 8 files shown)


evilebottnawi:
And it is not bugfix, it is feature, we never support `@value`s in selectors

jquense:
yup i'm thinking through the breaking change case, Might make this all opt-in initially...

Will definitely add tests

jquense:
I would call it a bugfix since the behavior is not supposed to be new, it's specced out by css-modules. That said, it's never worked and it adds new behavior so if you want to call it a feature i understand

evilebottnawi:
@jquense Let's do this step by step. Please clarify what each PR do:
1. https://github.com/css-modules/postcss-modules-scope/pull/4
2. https://github.com/css-modules/postcss-modules-values/pull/2

And how it is related to this PR

src/plugins/postcss-icss-parser.js
```diff
@@ -56,41 +56,45 @@ export default postcss.plugin(
           }
 
           const token = node.value;
-          const importIndex = imports[`$${token}`];
+          const replacement = importReplacements[token];
 
-          if (typeof importIndex === 'number') {
+          if (replacement != null) {
```
evilebottnawi: Why not `if (replacement) {}`?

src/plugins/postcss-icss-parser.js
```diff
@@ -9,22 +9,22 @@ export default postcss.plugin(
...
 
-      Object.keys(icss.icssImports).forEach((key) => {
+      let index = 0;
+      Object.keys(icssImports).forEach((key) => {
         const url = loaderUtils.parseString(key);
 
-        Object.keys(icss.icssImports[key]).forEach((prop) => {
-          const index = Object.keys(imports).length;
+        Object.keys(icssImports[key]).forEach((prop) => {
+          index += 1;
```
evilebottnawi: Why use `index`? Better solution will be good also

jquense: I was refactoring the existing code which used the `index` the same way (I believe) but it was calculating it on every loop iteration. e.g. `Object.keys(imports).length`. This way does less work, and seemed like a harmless refactor

src/plugins/postcss-icss-parser.js
jquense: it could be! The old want needed to check for `0` but that's not the case anymore so lets do that!

evilebottnawi: :+1:

src/plugins/postcss-icss-parser.js
evilebottnawi: :+1:

evilebottnawi:
Need fix CI problem with naming commit, please read CONTRIBUTING.md

evilebottnawi:
Also please accept CLA

jquense:
right so there are a few blocks to supporting this:

### postcss-modules-values PR: https://github.com/css-modules/postcss-modules-values/pull/2

This updates the values plugin so that it also replaces values in selectors. That behavior was always intended but it was using an old version of the icss utils that didn't support it.

### postcss-modules-scope PR: css-modules/postcss-modules-scope#4

Now that selectors are being replaced we run into a problem further down the chain in the scope plugin.

With the `values` PR this: 
```css
@value btn-class from './button.module.css'

.toolbar > .btn-class {
  margin-left: 10px
}
```

gets turned into:

```css
:import('./button.module.css') {
  1__const_btn-class-0: btn-class
}

.toolbar > .1__const_btn-class-0 {
  margin-left: 10px
}
```

So that's what we want, the _value_ of `btn-class` to be used as the class name here. **Unfortunately** the scope plugin with the default locals plugin will turn that selector into:

```css
:local(.toolbar) > :local(.1__const_btn-class-0) {
 ...
}
```

This leads to the value identifier (`1__const_btn`) being hashed turning it into something like `foo_module_12353-1__const_btn`, which we don't want, because it removes the usage of the import.

### This PR

The reason we need the last step is b/c _after_ that pipeline we need to replace instances of import identifiers with a webpack `require()` to the file and the identifier there. Previously webpack wasn't checking in selectors for instances of local alias's so we needed to fix that.

jquense:
Now that i'm thinking about it, it may be better to change the local by default plugin instead of the scope one...will check that

evilebottnawi:
Okey, we need major release for this feature, it is normal, but we need more tests here and in plugins

jquense:
yes I will add the tests, I wanted to make sure this wasn't going to be rejected out of hand before doing all that :)

evilebottnawi:
@jquense just update lock file in other PR: `rm -r node_modules && npm -r package-lock.json && npm install` and send a new PR with new lock file, when we rebase this PR (security problems in some deps)

evilebottnawi:
/cc @jquense PR looks good, but we have some big problem in many plugins and i think we should solve this problem first https://github.com/webpack-contrib/css-loader/issues/578 as patch release and when merge all PR as major release.

Problem:
`css-selector-tokenizer` break some selector (you can see this in first post in issue), we should migrate on `postcss-selector-parser` because he has better compatibility. Plugins which should be migrate on `postcss-selector-parser`:
- https://github.com/css-modules/postcss-modules-scope
- https://github.com/css-modules/postcss-modules-local-by-default

Can you help with this?

jquense:
sounds good i can work on that yeah 👍

evilebottnawi:
@jquense feel free to ping me

jquense:
ok updated the other PR, need that merged and released before I can easily write tests here!

jquense:
Sorry, didn't realize I was the bottleneck. Everything is up to date on my end. Waiting on https://github.com/css-modules/postcss-modules-local-by-default/pull/13 to be merged, should have been more clear, realized I just linked the issue

evilebottnawi:
@jquense Steps:
1. https://github.com/css-modules/postcss-modules-local-by-default/pull/13
2. https://github.com/css-modules/postcss-modules-scope/pull/4
3. This PR, right?

jquense:
only css-modules/postcss-modules-local-by-default#13
then this. I closed the other PR the local-by-default PR replaces it

evilebottnawi:
@jquense Just clarify, https://github.com/css-modules/postcss-modules-values/pull/2 need merge too?

jquense:
@evilebottnawi 

these both need to be merged first:
- https://github.com/css-modules/postcss-modules-values/pull/2
- https://github.com/css-modules/postcss-modules-local-by-default/pull/13

evilebottnawi:
/cc @jquense all plugins was released, before merge this we need do:
- Update postcss modules deps in this PR
- Add couple of tests

jquense:
will do 👍

jquense:
@evilebottnawi I'm running into a few issues with the existing tests and some of the escaped class names.

from `basic.css`
```
.\3A \`\( 
```

there is a space in the selector, when wrapped in a `:local()` block by the plugin it complains because spaces are not valid inside `:local()` selectors

jquense:
ah nevermind i didn't upgrade the scope package!

jquense:
ok @evilebottnawi Should be good to go. I didn't add test case b/c the existing css-modules test cases cover the change well already. See the diff for the snapshot

evilebottnawi:
@jquense we need couple test for smoke testing (ensure all deps in updated and feature works), no need deep testing just add one example

test/__snapshots__/modules-option.test.js.snap
```diff
@@ -5374,7 +5374,7 @@ Array [
   color: green;
 }
 
-._1tAbIwITRWAdZZE6wKNk9O {
+._3SRb8VDtr6fpZVzVUOJOM2 {
```
evilebottnawi: Very strange :confused:

test/__snapshots__/modules-option.test.js.snap
```diff
@@ -5362,7 +5362,7 @@ Array [
   color: blue;
 }
 
-._1wABXM_RabWHj--wsPrhvM {
+.green {
```
evilebottnawi: Need investigate why it is happens

jquense: its supposed to happen. this is an effect of the change, `other` is an imported value

jquense:
> we need couple test for smoke testing (ensure all deps are updated and feature works)

@evilebottnawi the tests we have _do_ this already, the changes in the existing snapshots are explicitly b/c of the changes we made

test/__snapshots__/modules-option.test.js.snap
evilebottnawi: @jquense let's rename selector, because it can be misleading (just add couple character)

evilebottnawi:
@jquense better create new test case for selectors (here https://github.com/webpack-contrib/css-loader/tree/master/test/fixtures/modules/tests-cases) and update snapshot, it is very easy and help to us

test/__snapshots__/modules-option.test.js.snap
jquense: i can, but this test is specifically testing the css-modules resolution logic. it really should include a resolution in a selector, I would consider the test incomplete otherwise, but if you want it to be a different test i can do that

evilebottnawi: @jquense, i agree with you, just want to avoid misleading for other developers which see PR and tests

test/fixtures/modules/values-in-selectors.css
```diff
@@ -0,0 +1,14 @@
+@value other from './values.css';
+@value button-class: btn;
+
+:import('imported-simple.css') {
+  other-child: imported-simple;
+}
+/* At the moment local-by-default  doesn't understand values .btn would get hashed */
```
evilebottnawi: hm, i think we should solve this before release new version

test/__snapshots__/modules-option.test.js.snap
```diff
@@ -1,5572 +1,53 @@
...
-	100% { left: 20px; }
-}
-
-.d1 {
-	animation: d1;
-	animation: d2, d3, d4;
-}
-
-.d2 {
-	animation: _d2;
```
evilebottnawi: Something wrong with snapshot

test/modules-option.test.js
```diff
@@ -54,6 +54,23 @@ describe('modules', () => {
     expect(stats.compilation.errors).toMatchSnapshot('errors');
   });
 
+  it.only('composes should values in selectors', async () => {
```
evilebottnawi: Remove `only`

test/fixtures/modules/values-in-selectors.css
jquense: I think it'd be worth deferring that. 

- it's not obviously wrong behavior (maybe not a bug)
- it's not simple to solve well
- it's not a common use case for this and easily worked around

evilebottnawi: @jquense hm, can you clarify with examples, i am afraid we can break many code with new release

jquense: I don't think this is a bug necessarily, and changing it would break _more_ people than this release currently will.

here is the current versions behavior:

```css
@value foo: blue;

// .foo become ._22ETk3ZhhvjVe-4dAULcsF
.foo { 
 display: block
}
```

That will stay the same. If explicitly want `.foo` to be replaced by `.blue` you need to do `:global(.foo)`. current code without `:global` won't change behavior

evilebottnawi: hm, based on https://github.com/css-modules/icss#import we should replace `foo` without `:global`, or maybe I misunderstood you. 

When modules were enabled we replaces values in selectors by default?

evilebottnawi: oh, i see, i think it is invalid behavior, when i am using `:global` i want selector will be global without any processing

jquense: `foo` isn't an `:import` (`other` is and is replaced) so i don't think so. To be clear what's happening is we are replacing `.foo` with `.blue`, and then `.blue` is hashed and exported.

> When modules were enabled we replaces values in selectors by default?

we replace _imports_ in selectors, that includes imported values not just imported classes. and yes it's on by default, since that's the correct behavior according to the ICSS spec


There are really two related but slightly different changes. `@value` replacement, which isn't defined by the spec. and proper `:import` replacement,  which is specced out.

the change we made to `local-by-default` is enable imports to be used in selectors without them breaking, but that plugin has no idea what a `@value` is, and probably shouldn't.

overall this is so complex because the css-module **default** is bad and involves a whole processing step to wrap classes in `:local` instead of making that the default already

jquense: > when i am using :global i want selector will be global without any processing

That's not clear to me. `global` just means "do not hash" it doesn't mean don't do any processing.

again tho this is really hard to change unless `local-by-default` plugins knows about `@value`s, you'd have to combine the two plugins for that, and that's a bit weird b/c `@value` isn't even a specced css-module thing, its just a plugin that css-loader happens to use

evilebottnawi: fair enough, we need update docs in this PR for other developers

jquense: awesome yeah i forgot about the docs. will update

evilebottnawi: :+1:

README.md
```diff
@@ -437,6 +437,46 @@ To import from multiple modules use multiple `composes:` rules.
...
+```
+
+```css
+@value button from './button.css';
+
+/* the 'button' class will be replaced with it's hashed value from 'button.css' */
+.toolbar > .button {
+  margin-left: 2rem;
+}
+```
```
evilebottnawi: Let's add example with at-rules like `@media` here and we can merge

evilebottnawi:
Also need fix CI

jquense:
@evilebottnawi this should be ready now 👍

jquense:
sure I can look into that

evilebottnawi:
@jquense thanks!

evilebottnawi:
Let's do release on this week

evilebottnawi:
/cc @jquense 
Find interesting problem:
```
@value v-primary: #BF4040;
@value s-black: black-selector;
@value m-large: (min-width: 960px);

.primary-selector {
  color: v-primary;
}

.s-black {
  color: black;
}

@media m-large {
  .header {
    padding: 0 20px;
  }
}
```

I think we should fix it (maybe better error message)

jquense:
what's happening that seems wrong with the example?

evilebottnawi:
@jquense error when you compile due this line `@value s-black: 'black-selector'; /* Problem here */` (quotes)

jquense:
I don't think this is a new issue? 

```css
@value color: 'black';

.s-black {
  color: color;
}
```

is possible now and also won't work. Agree that a nicer error would be good but I don't think that's in scope for this PR?

evilebottnawi:
@jquense I already implement supporting `@value` in selector (in master), thanks for PR, just find this error. Can you send other PR?

jquense:
Ya you bet 👍 should we close this one then if you've pulled it in otherwise 😁

evilebottnawi:
@jquense i keep open issue due we should fix problem what i describe above

evilebottnawi:
/cc @jquense i can't do release before we can't fix problem with:

```
@value v-string: 'my-content';

.v-string {}
```

jquense:
I'm sorry, i haven't had any time but i still don't understand why this is a blocker. The behavior of it not working is expected and consistent with the _current_ v2 behavior for value interpolation.

jquense:
Better, question what do you want it to do? having a selector with quotes is invalid, i don't think we should support that.

evilebottnawi:
Error like: `Error: string (\"'my-content'\") is not allowed in a :local block\` not just throw error with stack

jquense:
Unfortunately I don't think is no good place to check for and throw this error (unless you have an idea). It would likely involve a big refactor of the css-modules pipeline in order to even notice it since it's related to behavior across multiple postcss plugins.

I agree that the behavior isn't great, but it's not new behavior and it's LESS bad than other cases like [this](https://github.com/webpack-contrib/css-loader/pull/894#issuecomment-497086261) which fail silently. I don't think we should make it a release blocker.

Overall i'd be happy to work on these things generally but i don't have any access to the key repos or the authority to make larger structural changes necessary to fix these sort of systemic issues with css-modules. Besides yourself no one else is responsive on any repo either

evilebottnawi:
So we can't fix it in `postcss-modules-local-by-default`?

jquense:
I don't think so...you'd only catch the ones for imported values, so your example above would still fail. We could try and put it in the `values` plugin but i'm not sure we know when the replacement is for a selector, since they all use the utils. I'd say we should put it in the utils module but that might be a larger scope than we'd want?

evilebottnawi:
Maybe we just throw error on `String` node in `postcss-modules-local-by-default`? I think it is not difficult

evilebottnawi:
Solved in postcss-modules-local-by-default

evilebottnawi:
I think tomorrow will be release new major version

jquense:
awesome! glad you found a solution, I was thinking too hard about it 👍

PR #920, opened 2019-04-10
refactor: better export escaped locals

evilebottnawi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Better export locals for escaping selectors

### Breaking Changes

In theory yes becase it is change css output for classes and ids but previously implementation was wrong so it is bugfix

### Additional Info

No

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -7,6 +7,7 @@ import path from 'path';
 import cc from 'camelcase';
 import loaderUtils from 'loader-utils';
 import normalizePath from 'normalize-path';
+import cssesc from 'cssesc';
 
 /* eslint-disable line-comment-position */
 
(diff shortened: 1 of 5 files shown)


src/utils.js
```diff
@@ -79,16 +80,15 @@ function getLocalIdent(loaderContext, localIdentName, localName, options) {
...
-  // eslint-disable-next-line no-param-reassign
-  localIdentName = localIdentName.replace(/\[local\]/gi, localName);
-
-  const hash = loaderUtils.interpolateName(
-    loaderContext,
-    localIdentName,
-    options
-  );
-
-  return normalizeIdentifier(hash);
```
fa93hws: This line makes the replacing the `/` in path to `-` no longer working.
Is that on purpose?

PR #933, opened 2019-05-16
feat: add `resolve` option

### LINKED ISSUE #914: Migration strategy for adding tilde
There are two import syntax:

```css
@value foo from 'bar/baz.css';
```

```css
@value foo from '~bar/baz.css';
```

`css-loader@^1` tries to resolve 'bar/baz.css' in node_modules, and fails to resolve '~bar/baz.css'.
`css-loader@^2` tries to resolve 'bar/baz.css' only in local folder, and resolves '~bar/baz.css' in node_modules

## The problem

We have a few hundreds of react-components, which actively uses css imports. We cannot change all of them at once, because they are in different repositories, and some of them have cycled dependencies from each other. And it looks like there is no css-loader version which supports both formats.

Say, we have package A, which depends on B and C. Both, B and C have their own release cycle. If only one package updates its syntax, we will be unable to update that package until the second package also updates its syntax. Multiply that by few hundreds, and you will get our situation. So, we totally blocked.

What is an *official webpack strategy* for migration from 'bar/baz.css' to '~bar/baz.css' syntax?

niksy:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Closes #914.


evilebottnawi:
hm, `url` you already use as `Function` https://github.com/webpack-contrib/css-loader/#options, it is not solve problem? Maybe you can create minimum reproducible test repo?

niksy:
@evilebottnawi `url` expects boolean to be returned and tells css-loader should URL be handled or not. `resolve` solves cases like [3rd-party dependencies not being resolved from `node_modules`](https://github.com/webpack-contrib/css-loader/issues/914#issuecomment-479452105) and [migration strategy](https://github.com/webpack-contrib/css-loader/issues/914#issuecomment-486609889).

evilebottnawi:
@niksy if we modify `url` Function to allow return string and rewrite original url

niksy:
@evilebottnawi so, if `url` function returns boolean, it should tell css-loader to handle or not to handle specific URL, and if we return string, it should tell css-loader to use that string as new URL?

evilebottnawi:
@niksy yes

niksy:
@evilebottnawi I’ve moved functionality to `url` option, but I don’t know if this works properly for all cases. It seems like too much magic for one option.

ZauberNerd:
Hi @niksy @evilebottnawi, is this still being considered? We would need an option like this, because we have a lot of modules published to an internal registry that contain CSS modules and import CSS using `@import "@company/button/button.css";` or via `composes: foo from "@company/bar/baz.css";`.
In order to provide a smooth transition we would like to be able to customize the resolution algorithm and print deprecation warnings for paths that do not contain a tilde.

niksy:
@ZauberNerd I think maintainers need to decide which way this should go. IMO, adding new option is the most "secure" way since other solutions involve a lot of magic, but I could be wrong.

niksy:
Where are we with this? I think there is valid need for this since there are a lot of dependencies from npm which don’t follow "magic `~`" approach.

evilebottnawi:
@niksy you can disable url resolving or create postcss plugin, can you create minimum reproducible test repo, i need investigate situation to search better solution

yellow1912:
this is definitely a useful feature. I'm in the same situation where I need to customize the ways how the assets urls are resolved. Turning off url resolving from css-loader is the first step, but I still would like to be able to customize how it should be resolved.

ZauberNerd:
> @niksy you can disable url resolving or create postcss plugin, can you create minimum reproducible test repo, i need investigate situation to search better solution

@evilebottnawi does this help for reproducing? https://github.com/ZauberNerd/css-modules-repro
Actually, we (my team and I) are quite confused as to what behaviour with regards to CSS and module resolution is correct and we tried to make sense of it with the above linked repository, but unfortunately it seems all implementations and documentations are different, so we'll probably try to migrate our internal packages to match what this package is proposing (even though it differs from what the css modules spec/readme is suggesting).

evilebottnawi:
@ZauberNerd where you get infromation about url resolution for css modules? We do breaking changes and remove difference resolution between pure css and css modules

ZauberNerd:
@evilebottnawi the first sentence in the [css-modules/css-modules readme](https://github.com/css-modules/css-modules/blob/master/README.md#css-modules):
> All URLs (`url(...)`) and `@imports` are in module request format (`./xxx` and `../xxx` means relative, `xxx` and `xxx/yyy` means in modules folder, i. e. in `node_modules`).

evilebottnawi:
@ZauberNerd it is outdated information, you should prefer use information from this repository

evilebottnawi:
Don't forget you can create postcss plugin for migration, it is not hard

niksy:
I’m trying to understand why does css-loader needs `~` for modules resolution and it can’t follow NodeJS logic, which is defined in [CSS Modules readme](https://github.com/css-modules/css-modules/blob/master/README.md#css-modules). It seems like every tool follows that logic, except css-loader which forces developers to use convoluted solutions to consume anything that isn’t local to the project.

@evilebottnawi do you have any information regarding that?

evilebottnawi:
@niksy all loader in webpack use `~` for resolving modules from `node_modules`, not only css-loader. When you use 
```
url('something');
```
We can't determinate what is it? It is module or relative file. When you use pure CSS and use this code, css file load image/asset (regarding file) so we introduce `~` for loading from `node_modules`. css-loader use behavior as if you used pure CSS.

niksy:
@evilebottnawi 	

> ```
> url('something');
> ```
> 
> We can't determinate what is it? It is module or relative file.

If you use Node resolve algorithm, it’s module. Relative file would be `./something`. Forgive me if I don’t see some other reasons why Webpack chose `~` for resolving - probably because it’s defined in CSS spec which is valid reason, but that doesn’t cover `node_modules` stuff. Yes, you can probably override that with PostCSS plugin, but that seems like avoiding the original problem.

Maybe option should be to choose which resolving algorithm you want to use? By default, `~`, but you can opt-in to classic Node resolve with option?

evilebottnawi:
> If you use Node resolve algorithm, it’s module.

But is is violate how CSS work and resolve. It is css and it is code for browser, we can't use `node` resolution algorithm here.

What is problem using `~`? Please provide me example

niksy:
@evilebottnawi 	problem is perfectly described in [original post](https://github.com/webpack-contrib/css-loader/issues/914#issue-427257683).

evilebottnawi:
@niksy it is only for CSS modules, right?

niksy:
@evilebottnawi no, this affects both CSS Modules and classic approach.

egucciar:
Is it possible to do `@import '@local-modules/xyz.css'` ? We have a `resolve.modules` and within our local modules folder directory are doing `@` folders. This doesnt seem to work.

evilebottnawi:
Sorry for delay, fixed in the master

PR #941, opened 2019-05-28
feat: support `@value` at-rule in selectors

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Support `@value` at rule in selectors

### Breaking Changes

Yes

BREAKING CHANGE: if you have name of `@value` in selector it will be replaced, you need to rename your name of `@value` or rename your selector

### Additional Info

No

For next major release we need:
- fix `nothing` in test of css-loader
- fix problem when you use `@value test: 'something'` in selector
- refactor postcss-icss plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,33 @@ To import from multiple modules use multiple `composes:` rules.
 }
 ```
 
+##### `Values`
+
+You can use `@value` to specific values to be reused throughout a document.
+
+We recommend use prefix `v-` for values, `s-` for selectors and `m-` for media at-rules.
+
+```css
+@value v-primary: #BF4040;
+@value s-black: black-selector;
+@value m-large: (min-width: 960px);
+
+.header {
+  color: v-primary;
+  padding: 0 10px;
+}
+
+.s-black {
+  color: black;
+}
+
+@media m-large {
+  .header {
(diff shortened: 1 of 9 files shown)


Hypnosphi:
Is there any way to disable this feature?

evilebottnawi:
@Hypnosphi no, just rename your variable to `v-myClass`

> I use the following pattern a lot, so now I need a large refactoring to make it work again

It is breaking change for `css-loader@3`, so it is normal when you need refactor

Hypnosphi:
> rename your variable to `v-myClass`

That would mean that in `some/other.js` i would have to use something like `styles['v-myClass']` which looks suboptimal comparing to `styles.myClass`

evilebottnawi:
@Hypnosphi just use other class or change variable name, it can't be opt, sorry

princed:
@evilebottnawi Would you accept a PR making it possible to opt-out? In our project, we'd have to update about 5000 lines because of the feature.

evilebottnawi:
@princed it can be options, postcss plugins were rewritten

Issue #862, opened 2018-12-07
[2.0.0] getLocalIdent validation


gpoitch:
* Operating System: macOS 10.14.1
* Node Version: 11.2.0
* NPM Version: 6.4.1
* webpack Version: 4.27.1
* css-loader Version: 2.0.0

### Expected Behavior

Previously, if `getLocalIdent` returned a falsey value, css-loader would fallback to using `localIdentName`

### Actual Behavior

css-loader throws an error:
```
ValidationError: CSS Loader Invalid Options
options.getLocalIdent should pass "instanceof" keyword validation
```

### Code

```js
const isProduction = false

const cssLoader = {
  loader: 'css-loader',
  options: {
    modules: true,
    localIdentName: '[local]__[hash:base64:3]',
    getLocalIdent: isProduction && function() {
       return 'foo' // other logic here, but just return a string for purpose of issue
    }
  }
}
```

### How Do We Reproduce?

Run tests with those loader options

evilebottnawi:
@gpoitch what you expected from `getLocalIdent` when you set it `false`?

gpoitch:
I expect it to fallback to using `localIdentName`.
Worked around it by explicitly setting to `undefined` instead of a falsey value.
```js
getLocalIdent: isProduction ? function(){ /*...*/ } : undefined
```

evilebottnawi:
@gpoitch maybe we can accept `false` value for easy configuration, what do you think?

gpoitch:
@evilebottnawi yes, that'd be nice, as it worked in 1.x.  Otherwise that is a tiny breaking change

evilebottnawi:
@gpoitch let's mark it as bug and fix in `2.0.1`

evilebottnawi:
previous we don't have schema options and tests so i can lose something, sorry

PR #963, opened 2019-06-28
feat: allow `modules.getLocalIdent` to return a falsy value

ziir:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

With the suggested changes, `modules.getLocalIdent` would be allowed to return a falsy value, and in such case, the default `getLocalIdent` would be used instead.

I am setting up a new project with `webpack`, I wish to apply the same chain of loaders on different kinds of `.scss` files:
- `main.scss` specified as an `entry` in my configuration, meant to be used as global stylesheet
- `App.scss` SCSS modules which are imported from their matching component such as `App.js`

This setup works very well for me, except that `main.scss` classnames necessarily go through `css-loader` with `modules` enabled and using `localIdentName`.

See the following example:
```js
{
  test: /\.scss$/,
  loader: [
    MiniCSSExtractPlugin.loader,
    {
      loader: 'css-loader',
      options: {
        importLoaders: 1,
        modules: {
          localIdentName: '[name]-[local]__[hash:base64:5]',
        },
      },
    },
    {
      loader: 'sass-loader',
    },
  ],
},
```

While browsing issues, looking for a way to not use `localIdentName` for my `main.scss` file, I stumbled upon #862.
`getLocalIdent: () => foo && 'bar'` seemed to be a great fit for my use-case.
The behavior introduced in #865 wouldn't have worked for me though, here's what I did instinctively:
```js
modules: {
  localIdentName: '[name]-[local]__[hash:base64:5]',
  getLocalIdent: (context, localIdentName, localName) =>
    context.context.includes('shared/assets/styles') && localName,
},
```
Note: my `main.scss` file resides in `shared/assets/styles`.
This way, my SCSS modules will use `localIdentName` and `main.scss` will use `localName`.

### Breaking Changes

None

### Additional Info

Thank you for this great package :)
I'm willing to make the final changes if this looks reasonable, thank you in advance for any response.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2413,22 +2413,22 @@
       "dev": true
     },
     "cacache": {
-      "version": "11.3.2",
-      "resolved": "https://registry.npmjs.org/cacache/-/cacache-11.3.2.tgz",
-      "integrity": "sha512-E0zP4EPGDOaT2chM08Als91eYnf8Z+eH1awwwVsngUmgppfM5jjJ8l3z5vO5p5w/I3LsiXawb1sW0VY65pQABg==",
+      "version": "11.3.3",
+      "resolved": "https://registry.npmjs.org/cacache/-/cacache-11.3.3.tgz",
+      "integrity": "sha512-p8WcneCytvzPxhDvYp31PD039vi77I12W+/KfR9S8AZbaiARFBCpsPJS+9uhWfeBfeAtW7o/4vt3MUqLkbY6nA==",
       "dev": true,
       "requires": {
-        "bluebird": "^3.5.3",
+        "bluebird": "^3.5.5",
         "chownr": "^1.1.1",
         "figgy-pudding": "^3.5.1",
-        "glob": "^7.1.3",
+        "glob": "^7.1.4",
         "graceful-fs": "^4.1.15",
         "lru-cache": "^5.1.1",
         "mississippi": "^3.0.0",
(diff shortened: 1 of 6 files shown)


evilebottnawi:
Please accept CLA

evilebottnawi:
Also need update https://github.com/webpack-contrib/css-loader/blob/master/src/options.json#L41 and add test

ziir:
@evilebottnawi Thanks, signed the CLA.

I think I'm missing something, `localIdentName` remains a `string` and I cannot update `getLocalIdent` function return type?

I'm also not sure why some buils are failing.

evilebottnawi:
i was wrong, need update snapshots `npm run test:only -- -u`

ziir:
I already have updated the snapshots:
https://github.com/webpack-contrib/css-loader/pull/963/commits/bca916378eead10bc0a01952b27669da97414d7a#diff-3cb22b4647b22161352a612896f9eaa8

evilebottnawi:
@ziir update `webpack` to latest version in `package.json` and `package-lock.json` and run updating again

ziir:
@evilebottnawi thnx, all checks are green now :)

evilebottnawi:
Thanks, release will be on next week (Monday i think)

PR #968, opened 2019-07-15
feat: minify classNames

Yegorich555:
This PR contains a:

- [ ] **bugfix**
- [ x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Reducing classNames is must have improvement of the performance for production build. It will be very helpful using such option for changing classNames to 'a_b' like it Google does.
Now for using this we need to set option
```javascript
options: {
    modules: {
        minifyNames: true
    },
}
```

### Breaking Changes

There is no breaking changes at all. Only added option 'minifyNames, package 'incstr' (helps with generating unique ids) and some functions in utils.js.

### Additional Info
I guess that the provided feature can be improved. If you aren't satisfied this, could you implement own, please?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5079,7 +5079,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -5100,12 +5101,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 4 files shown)


evilebottnawi:
No need this, it is out of scope css-loader, also solution is invalid

evilebottnawi:
use own `getLocalIdent` implementation if you need this

Yegorich555:
Thanks for attention. A lot of people and I use own implementation for this purpose. I wish only to set a single solution for this. And I don't understand why is it out of scope css-loader? Since only css-loader implements css-modules option I don't see any ways to implement this without css-loader.

evilebottnawi:
>  A lot of people and I use own implementation for this purpose

Yes, because we can cover all 100% cases

> I don't understand why is it out of scope css-loader

Because for other developers your solution can break code, this is applicable only for your project

Yegorich555:
About the invalid solution: yes, now I see that, but the latest commit from the master branch failed 'test' also. I guess it not my deal at all since I don't support this package and I don't know enough what brokes tests. But I just wanted to provide a feature for minimizing classNames. It would be very helpful if in the future you would add such a useful rule.
In any case, I appreciate your attention. I will implement my own npm-package for getLocalIdent, but unfortunately, it is a crutch and it's not an elegant solution for this case.

evilebottnawi:
> it is a crutch and it's not an elegant solution for this case

Yep, but as said above we can't implement `elegant solutions` for all cases so we provide for developers options

PR #973, opened 2019-07-17
test: escaped in selector

### LINKED ISSUE #966: values in selectors aren't escaped properly.
on latest css-loader. I'd expect this to be true for any defined value but not importing classes where the local ident is generated.

// button.css
```css
.button {

}
```
// toolbar.css
```css
@value btn from './button.css'

.toolbar > btn {
	color: red
}
```

```js
rules: [
  { 
  test: /\.css/, 
  use: {
    loader: 'css-loader',
    options: {
		modules: {
			getLocalIdent: (ctx, name) => `${name}.hey`
		}
	}
  }
]
```

The local names would need escaping and webpack outputs:

```
.button\.hey {
}
```

which is correctly escaped but in toolbar its

```css
.toolbar\.hey > .button.hey {
}
```

note that the interpolated value isn't escaped. I'm not sure why this is happening maybe it needs to be re-escaped?

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #966

### Breaking Changes

No

### Additional Info

/cc @jquense looks all fine, can't see problems or bugs

### DIFF
--- a/test/__snapshots__/modules-option.test.js.snap
+++ b/test/__snapshots__/modules-option.test.js.snap
@@ -6355,6 +6355,33 @@ exports.locals = {
 
 exports[`modules issue #861: warnings 1`] = `Array []`;
 
+exports[`modules issue #966: errors 1`] = `Array []`;
+
+exports[`modules issue #966: module (evaluated) 1`] = `
+Array [
+  Array [
+    1,
+    ".button.hey {
+  color: red;
+}
+",
+    "",
+  ],
+]
+`;
+
+exports[`modules issue #966: module 1`] = `
+"exports = module.exports = require(\\"../../../../src/runtime/api.js\\")(false);
+// Module
+exports.push([module.id, \\".button.hey {\\\\n  color: red;\\\\n}\\\\n\\", \\"\\"]);
+// Exports
+exports.locals = {
+	\\"button\\": \\"button.hey\\"
+};"
+`;
+
+exports[`modules issue #966: warnings 1`] = `Array []`;
+
 exports[`modules issue #967: errors 1`] = `Array []`;
 
 exports[`modules issue #967: module (evaluated) 1`] = `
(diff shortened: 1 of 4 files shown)


test/__snapshots__/modules-option.test.js.snap
```diff
@@ -6355,6 +6355,33 @@ exports.locals = {
 
 exports[`modules issue #861: warnings 1`] = `Array []`;
 
+exports[`modules issue #966: errors 1`] = `Array []`;
+
+exports[`modules issue #966: module (evaluated) 1`] = `
+Array [
+  Array [
+    1,
+    ".button.hey {
```
jquense: the bug is right here: `.button.hey` is very different than: `.button\.hey`. The former is a selector for when both `.button` and `.hey` are on an element

jquense:
The test case here is not accurate. It illustrates the bug but doesn't contain an assert checking for it

evilebottnawi:
@jquense what you expected?

evilebottnawi:
Let's move discussion in https://github.com/webpack-contrib/css-loader/issues/966

test/__snapshots__/modules-option.test.js.snap
```diff
@@ -6355,6 +6355,33 @@ exports.locals = {
...
+]
+`;
+
+exports[`modules issue #966: module 1`] = `
+"exports = module.exports = require(\\"../../../../src/runtime/api.js\\")(false);
+// Module
+exports.push([module.id, \\".button.hey {\\\\n  color: red;\\\\n}\\\\n\\", \\"\\"]);
+// Exports
+exports.locals = {
+	\\"button\\": \\"button.hey\\"
```
jquense: This is also wrong, but if you escaped in `getLocalIdent` it would be fixed

PR #974, opened 2019-07-18
test: clearer test case for 966

jquense:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Follow up to #973


jquense:
Let me try and figure out what is different about this vs my setup @evilebottnawi

evilebottnawi:
I see problem, when you use `localIdentName` we escape all special CSS characters, but when you use `getLocalIdent` we output as returned value without escaping, maybe you are right and it is bug, i think we should use the same logic for both cases

evilebottnawi:
We escape all characters in placeholder excluding `[locals]`

jquense:
AHHHH you're right! That's the problem it assumes that the local name is _already_ css safe. I think it's safe to make that change? Or would it not be ok?

evilebottnawi:
@jquense i think we should always escape non standard CSS character in both cases, why? because you path can contains invalid characters for CSS, `localIdentName` already works as expected, but getLocalIdent - no

PR #1005, opened 2019-12-05
fix: check for `url` being null before accessing `__esModules` property

charlessuh:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Perhaps this is an unusual use case, but I'm using a custom final loader which can export `null` (i.e. the source is `module.exports = null`), e.g. if it determines a module will not be needed.

I suppose I could export an empty object, but in some cases I need to check if a module was removed in this manner, and it's easier to check for `null` which is falsy (as opposed to an empty object which is truthy.)

### Breaking Changes
n/a

### Additional Info
n/a

### DIFF
--- a/src/runtime/getUrl.js
+++ b/src/runtime/getUrl.js
@@ -5,7 +5,7 @@ module.exports = (url, options) => {
   }
 
   // eslint-disable-next-line no-underscore-dangle, no-param-reassign
-  url = url.__esModule ? url.default : url;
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Need test

evilebottnawi:
What is use case using null? Can you provide example?

PR #1008, opened 2019-12-09
feat: support `pure` css modules

### LINKED ISSUE #985: Add support for pure CSS Modules
- Operating System: N/A
- Node Version: N/A
- NPM Version: N/A
- webpack Version: N/A
- css-loader Version: `3.2.0`

### Feature Proposal

Ability to use `modules: 'pure'`, a very-desirable CSS Modules mode.

### Feature Use Case

Defining global CSS using CSS Modules is arguably incorrect behavior, which `modules: true` / `modules: 'local'` allows.

The underlying `postcss-modules-local-by-default` library has a more desirable mode called `pure`, which only allows a `:global` selector when combined with `:local`. 
This ensures users are not defining global effects within a CSS Modules file.

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #985

### Breaking Changes

No

### Additional Info

No


Timer:
Thank you so much! I was actually about to send a PR for this today, as we needed it!

PR #1040, opened 2020-01-10
fix: do not duplicate css on `composes`

### LINKED ISSUE #1037: "Composes" duplicating css (again)
- webpack Version: 4.19.1
- css-loader Version: 3.4.1

It's copy of #772 but with latest css-loader.

### Expected Behavior
It shouldn't duplicate css class.

### Actual Behavior
When a class name within a css file is shared using "composes" in a 2 degree level, it gets duplicated (when each class has its own file).

For example (each class has its own css file):

Case 1 (composes in 1 degree level)
"primary-button" which composes "button" and;
"secondary-button" which composes "button".
This case is ok, and nothing gets duplicated.

Case 2 (composes in 2 degree level)
"next-button" which composes "primary-button" which composes "button" and;
"back-button" which composes "secondary-button" which composes "button".
On this case, the "button" class gets duplicated.

The second case was tested with the following plugins:
style-loader (see print): https://raw.githubusercontent.com/ivensgoncalves/Webpack_StylesDuplication/master/prints/style-loader.png
mini-css-extract-plugin (see print): https://raw.githubusercontent.com/ivensgoncalves/Webpack_StylesDuplication/master/prints/mini-css-extract-plugin.png

On both plugins the "button" class got duplicated.

### Code
With latest css-loader:
https://github.com/vlazh/Webpack_StylesDuplication

### How Do We Reproduce?
A minimal reproducible project with this issue:
https://github.com/vlazh/Webpack_StylesDuplication

To test with style-loader, execute "npm run start1"
To test with mini-css-extract-plugin, execute "npm run start2"

evilebottnawi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #1037

### Breaking Changes

No

### Additional Info

No


jquense:
thanks for all the hard work @evilebottnawi !

evilebottnawi:
@jquense I will ping you when I done release, check if everything is well fixed

src/runtime/api.js
```diff
@@ -22,15 +22,33 @@ module.exports = function(useSourceMap) {
 
   // import a list of modules into the list
   // eslint-disable-next-line func-names
-  list.i = function(modules, mediaQuery) {
+  list.i = function(modules, mediaQuery, dedupe) {
```
fa93hws: Is that possible to have it to be `true` by default?
e.g.
`function(modules, mediaQuery, dedupe = true)`?

evilebottnawi: Why?

PR #1044, opened 2020-01-22
fix: deduplicate css module for @import

fa93hws:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
`composes` is fixed in https://github.com/webpack-contrib/css-loader/pull/1040/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R329 but `@import` isn't.
The use-case can be refereed to the added test case.

"source" which **imports** "black" which composes "color" and;
"source" which **imports** "swhite" which composes "color".
On this case, the "color" module gets duplicated.

### Breaking Changes
No

### Additional Info
N/A


test/__snapshots__/import-option.test.js.snap
```diff
@@ -616,118 +598,6 @@ Array [
...
-    "screen and (orientation:landscape)",
-  ],
-  Array [
-    "../../src/index.js?[ident]!./import/test.css",
-    ".test {
-  a: a;
-}
-",
-    "screen and (orientation:landscape)",
-  ],
```
fa93hws: They are de-duplicated.

test/__snapshots__/import-option.test.js.snap
```diff
@@ -755,36 +625,50 @@ Array [
...
     "./import/import.css",
     "@import url(http://example.com/style.css);",
+    "",
+    true,
   ],
   Array [
     "./import/import.css",
     "@import url(http://example.com/style.css);",
+    "",
+    true,
```
fa93hws: Do they need to be de-duplicated? It's the `url` case.
If the de-duplication is necessary in all cases, probably we should change the runtime https://github.com/webpack-contrib/css-loader/pull/1040#discussion_r369365652.

evilebottnawi:
`@import` should not be deduplicate, it is invalid, please open issue with reproducible test repo

Issue #994, opened 2019-11-11
Auto resolving css-modules import and default import


Yegorich555:
### Feature Proposal
We have two imports ES6 and It looks smarter when resolving some imports will be work in the following automatic way:
```import styles from 'someFiles.css'``` as css-module 
```import 'someFiles2.css'``` as default import (without css-module-specification)

### Feature Use Case
The main idea is simplifying webpack config and avoiding misunderstanding in mixed project (when we have some files as css-modules and some-files as default-css).

Is implementing of this possible?

evilebottnawi:
`import from 'someFiles2.css'` invalid JS syntax, so it is impossible

evilebottnawi:
Closing due to inactivity.

Yegorich555:
How soon can it be implemented

Yegorich555:
Sorry. I made mistake. Fixed in the first message:
```import styles from 'someFiles.css'``` as css-module 
```import 'someFiles2.css'``` as default import (without css-module-specification)

evilebottnawi:
@Yegorich555 can you clarify what you want, hard to udnestand

Yegorich555:
According to css-modules specification we have to use:
``` import styles from 'someFile'``` and use **styles** as object.

But for switching on css-modules rule we should configure this for exact-files (by regex pattern). Does it makes sense? Why can not it be resolved automatically? I'm not sure is it question for css-loader?
The different imports can provide different rules to us. And It seems can be resolved these without configuring css-loader at all.
```import styles from 'someFile.css'``` it can be resolved as css-module
```import 'someFile.css'``` it can be resolved as default css-import (without css-module-specification)

Why do I need this:

1. Avoiding the overwhelmed webpack configuration.
2. Such auto-resolving imports excludes headache in projects when we have some files like css-module and some files as plain-css (that should be only attached to the project)

Here you can see 2 config sections which I use in projects:
``` js
 {
    // rules for style-files
    test: /\.css$|\.scss$/,
    oneOf: [
      /* config .oneOf('normal-modules') - rule for [name].module.css files - rule for css-modules */
      {
        test: /\.module\.\w+$/,
        use: [
          isDevServer
            ? "style-loader" // it extracts style directly into html (MiniCssExtractPlugin works incorrect with hmr and modules architecture)
            : MiniCssExtractPlugin.loader, // it extracts styles into file *.css
          // TODO: improve plugin for splitting by files for dev purpose
          {
[...]

PR #1046, opened 2020-01-22
feat: export default `getLocalIdent`

niki-timofe:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

For example, use default `getLocalIdent`, but override it's behavior for specific files:

```js
const getLocalIdent = require('css-loader/dist/utils').getLocalIdent;

...

            {
                loader: 'css-loader',
                options: {
                    modules: {
                        localIdentName: '[local]--[hash:base64:5]',
                        getLocalIdent: (context, localIdentName, localName, options) => getLocalIdent(
                            context,
                            context.resourcePath.endsWith('.modules.scss')
                                ? localIdentName
                                : '[local]',
                            localName,
                            options,
                        ),
                    },
                },
            }
```

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -74,7 +74,7 @@ function getLocalIdent(loaderContext, localIdentName, localName, options) {
       .replace(reRelativePath, '-')
       .replace(/\./g, '-'),
     { isIdentifier: true }
-  ).replace(/\\\[local\\\]/gi, localName);
+  ).replace(/\\\[local\\]/gi, localName);
 }
 
 function normalizeUrl(url, isStringValue) {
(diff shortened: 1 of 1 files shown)


niki-timofe:
I understand. May be adding option for filtering files will be better solution?

evilebottnawi:
Which files?

niki-timofe:
As in example, I've bunch of legacy files, which not supposed to use css-modules.

evilebottnawi:
Other good solution what we can implement https://github.com/webpack-contrib/css-loader/issues/994

evilebottnawi:
We can implement that right now (`test` option should be false in current version, but we can set it to `true` in next release)

niki-timofe:
Oh, it's cool. I'll try to help implement that.

PR #1049, opened 2020-02-05
feat: accept semver compatible postcss AST

jquense:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

This is a nice optimization css-laoder does, but seems hard to hit, given how little control the end user has to ensure they only have one particular postcss.


src/index.js
```diff
@@ -1,25 +1,25 @@
...
 /*
   MIT License http://www.opensource.org/licenses/mit-license.php
   Author Tobias Koppers @sokra
 */
 import validateOptions from 'schema-utils';
-import postcss from 'postcss';
-import postcssPkg from 'postcss/package.json';
-
-import { getOptions, isUrlRequest } from 'loader-utils';
+import satisfies from 'semver/functions/satisfies';
```
evilebottnawi: Let's use public API, no need hardcore path to function, it can be changed in theory

jquense: this is public https://www.npmjs.com/package/semver The readme, points to this for smaller footprints

package.json
```diff
@@ -54,7 +54,8 @@
     "postcss-modules-scope": "^2.1.1",
     "postcss-modules-values": "^3.0.0",
     "postcss-value-parser": "^4.0.2",
-    "schema-utils": "^2.6.0"
+    "schema-utils": "^2.6.0",
+    "semver": "^7.1.2"
```
evilebottnawi: https://github.com/npm/node-semver/blob/master/package.json#L36

I think we need to use lower version

jquense: I can try and find one. Ya'll know that node 8 is End of Life right?

evilebottnawi: We still support node@8 https://github.com/webpack-contrib/css-loader/blob/master/package.json#L16, In the near future there will be a release that will throw out support node@8, but right now it is breaking change, so we should avoid it

jquense: yup, just wanted to make sure ya'll knew.

PR #1053, opened 2020-03-18
work-around for lost paths to source files in source map

alexhisen:
This PR contains a:

- [X] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

When css-loader calls postcss with an existing source map previously generated by postcss-loader, the resulting source map ends up with just the file name for sources without any paths. If there are then collisions of this file name with other same-named files in other directories, only one of these source maps ends up being retained.

This is particularly noticeable for the react-toolbox project which has about a dozen files all named theme.modules.css in different directories.

See https://github.com/alexhisen/mobx-forms-demo project for example config, etc. Note it's currently set up to use this fork, so it doesn't exhibit the problem any more.

### Additional Info

This fix is a dirty hack workaround. There is probably a better way to fix this issue but I don't know enough about webpack, postcss, or this loader to know how to fix this better.


src/index.js
```diff
@@ -87,6 +89,15 @@ export default function loader(content, map, meta) {
         : false,
     })
     .then((result) => {
+      // work-around for apparent bug in postcss that loses the paths to source
+      // files in existing map previously generated by postcss-loader:
+      if (result.map && map && map.file) {
+        /* eslint-disable no-underscore-dangle, no-param-reassign */
+        result.map._sourceRoot = path.dirname(map.file);
```
evilebottnawi: Ideally `sourceRoot` should be `empty` and each sources relative to webpack context with unique prefix, we will improve source maps for css pipeline in near future, so please open issue with reproducible test repo and we will find solution

evilebottnawi:
Sorry for delay, let's close this in favor https://github.com/webpack-contrib/css-loader/pull/1169, we improve source map supporting, the next version will fix most of bugs

Issue #1028, opened 2019-12-19
Reduce bundle size via one-letter css classname hash strategy


denisx:
A worked code sample for https://dev.to/denisx/reduce-bundle-size-via-one-letter-css-classname-hash-strategy-10g6

Improving bundle compression to 40% of filesize via change standard css classname hash for splitting to one-letter name strategy and filepath.

webpack
```
// at header
const OneLetterCss = require('../utils/one-letter-css');
const MyOneLetterCss = new OneLetterCss();

// config
module: {
        rules: [
             {
                test: /\.css$/,
                use: [
                    {
                        loader: 'css-loader',
                        options: {
                            sourceMap: true,
                            modules: {
                                localIdentName: '[hash:base64:8]',
                                getLocalIdent: MyOneLetterCss.getLocalIdent
...
```

class 
```
/**
 * @author denisx <github.com@denisx.com>
 */

const loaderUtils = require('loader-utils');

module.exports = class OneLetterCss {
  constructor() {
    // Save char symbol start positions
    this.a = 'a'.charCodeAt(0);
    this.A = 'A'.charCodeAt(0);
    // file hashes cache
    this.files = {};
    this.lastUsedFiles = -1;
    this.lastUsedClasses = -1;
    /** encoding [a-zA-Z] */
    this.symbols = 52;
    /** a half of encoding */
    this.half = 26;
    /** prevent loop-hell */
    this.maxLoop = 10;
  }

  /** encoding by rule count at file, 0 - a, 1 - b, 51 - Z, 52 - ba, etc */
  getName(lastUsed) {
    const { a, A, symbols, maxLoop, half } = this;
    let name = '';
    let loop = 0;
    let main = lastUsed;
    let tail = 0;

    while (
      ((main > 0 && tail >= 0) ||
        // first step anyway needed
        loop === 0) &&
      loop < maxLoop
    ) {
      const newMain = Math.floor(main / symbols);

      tail = main % symbols;
      name = String.fromCharCode((tail >= half ? A - half : a) + tail) + name;
      main = newMain;
      loop += 1;
    }

    return name;
  }

  getLocalIdent(context, localIdentName, localName) {
    const { resourcePath } = context;
    const { files } = this;

    // check file data at cache by absolute path
    let fileShort = files[resourcePath];

    // no file data, lets generate and save
    if (!fileShort) {
      this.lastUsedFiles += 1;

      // if we know file position, we must use base52 encoding with '_'
      // between rule position and file position
[...]

PR #1062, opened 2020-03-29
feat: resolve issue #1028: optimising naming for best compression

denisx:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

resolve issue #1028 

For efficient gzip/br compression, plugin combine css hash via one symbol name,
as a classname position at file, with filepath `hash:base64:8`, to have strong sequences.


denisx:
have some bugs from master :/ 
npm audit fix change pakage-lock.json ...

joseprio:
We use a similar technique, but just defining a custom `getLocalIdent`. However, I don't think this plays well with the caching mechanism, as the classname generation isn't stable, right?

denisx:
@joseprio 
> However, I don't think this plays well with the caching mechanism, as the classname generation isn't stable, right?
Css-loader get async files and parse, it is allways randomized work. Class save classnames positions for every file (a,b,c,...). As a file-path, it is always stable content hash for project. It is cache frendly.

denisx:
and https://gist.github.com/denisx/9dc0e8c542b810e67f41691c89e7b948 for check hash length

denisx:
Add css hash len suggest - we have full hash map after css-1-letter running.

PR #1083, opened 2020-04-30
Add option for alternative postcss syntaxes

mismosmi:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

My project exports raw scss in order to apply user-themeing through scss variables after the main bundle has been compiled. Therefore I want to use postcss-loader > css-loader > MiniCssExtractPlugin without using sass-loader.

As css-loader is built around postcss this is easily possible if there is an option to just pass a postcss syntax (or even just the parser) to the postcss instance in css-loader. 

### Additional Info

It does not really matter to me if this option is named syntax or if it is some object called postcss-options that takes a syntax-parameter or, you know, whatever. I just need to get css-loader to work on scss.


mismosmi:
Okaay, I have no idea what's going on on ubuntu/node8/webpack next but I don't see how it would be related to my changes. If you think it is please let me know and I'll see what I can do about it.

evilebottnawi:
Sorry, we don't need this option here, you should use `postcss-loader`

mismosmi:
Well, but I need css-loader to extract locals (`:export`) and to handle `url`s.

I guess I'll have to maintain a fork then :(

evilebottnawi:
You can use css-loader under your loader if you need, can you provide full example of case, better simple reproducible test repo with examples in readme, I am ready to watch it and give tips on how to organize it better.

mismosmi:
Here's a full example of what I'm trying to do:

https://github.com/mismosmi/css-loader-scss-example

I need to extract the scss to process later on with user-set variables while passing the `:export`s (including the class name mapping from css-modules) on to the js.

mismosmi:
Aaand in the example repo it works just fine while in my production repo I get `CssSyntaxError` in css-loader.

I'll make a wild guess and assume that in the example repo css-loader takes the already-parsed AST from postcss loader and in the prod repo it doesn't for some reason?

evilebottnawi:
`postcss-modules` is very old plugin and unmaintained, please do not use it

evilebottnawi:
Why do you need this pipeline?

mismosmi:
Uh, if I can use css-loader with scss I don't have to use postcss-modules. That's really cool.

I'm sitting on this large-ish codebase that's worked with raw css class names so far and now we're running into namespacing issues and want to start using css modules.
We also need to extract scss from the app, not the already compiled css, because in a later step we override scss variables with user values for themeing. Therefore we cannot use sass-loader between postcss-loader and css-loader as recommended.

But as it seems right now if I can get css-loader to just use the AST from postcss loader I won't even need this changed so I won't have to waste more of your time.

PR #1067, opened 2020-04-03
feat: auto resolving css-modules import

cap-Bernardito:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

fixes: #994 

### Breaking Changes

no

### Additional Info

no

PR #1086, opened 2020-05-28
feat: allow modules.auto to be a filter function

BPScott:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

The modules.auto feature as added in #1067 is pretty sweet. but sometimes you need more than a simple regex to check against your file path.

This PR allows `modules.auto` to be a function that that accepts the resourcePath and returns a boolean.

Along the way I tweaked some test case names to be a little clearer about what gets passed into modules.auto, which causes a bit of snapshot diff noise.

### Breaking Changes

n/a

### Additional Info

n/a

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -117,6 +117,10 @@ function shouldUseModulesPlugins(modules, resourcePath) {
     return modules.auto.test(resourcePath);
   }
 
+  if (typeof modules.auto === 'function') {
+    return modules.auto(resourcePath);
+  }
+
   return true;
 }
 
--- a/README.md
+++ b/README.md
@@ -542,7 +542,7 @@ module.exports = {
 
 ##### `auto`
 
-Type: `Boolean|RegExp`
+Type: `Boolean|RegExp|Function`
 Default: `'undefined'`
 
 Allows auto enable css modules based on filename.
@@ -576,7 +576,7 @@ module.exports = {
 
 ###### `RegExp`
 
-Enable css modules for files based on a filename and satisfying your regex.
+Enable css modules for files based on the filename satisfying your regex check.
 
 **webpack.config.js**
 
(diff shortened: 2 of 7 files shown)


src/utils.js
```diff
@@ -117,6 +117,10 @@ function shouldUseModulesPlugins(modules, resourcePath) {
     return modules.auto.test(resourcePath);
   }
 
+  if (modules.auto instanceof Function) {
+    return modules.auto(resourcePath);
+  }
```
evilebottnawi, line 122: Please use `typeof`

BPScott:
@evilebottnawi: Thanks! I've updated it to use typeof and have added some validate-options tests

BPScott:
CI looks unrelated. It's failing running webpack 5 tests on node 8 thanks to BigInt not existing. That seems reasonable as they bumped their minimum version a while back: https://github.com/webpack/webpack/pull/9868

Sounds like we should remove wepback 5 in node 8 from the test matrix?

evilebottnawi:
@BPScott Yes, let's test webpack@5 only for 10/12/14

BPScott:
#1089 Removes that CI check. I'll rebase this PR once that is merged.

evilebottnawi:
@BPScott merged :+1:

evilebottnawi:
Thanks!

TheMallen:
Any idea when this will be released? I see that it hasn't been yet.

evilebottnawi:
ETA is the next week

TheMallen:
Thanks!

TheMallen:
@evilebottnawi is there any process we could help with to help get the updates released?

evilebottnawi:
@TheMallen Make 48 hours in one day :smile: Release will be today

Issue #1134, opened 2020-07-27
v4 breaks css-module-loader 


jquense:
Hey @evilebottnawi the change that only adds the ICSS plugins breaks the strategy used by https://github.com/jquense/css-module-loader to convert modules into `:import` and `:export` rules which depended on the (admittedly weird) behavior of css-loader always including the icss plugin.

It's not enough to turn on `modules` tho b/c css-loader will then rehash everything. For this sort of thing to work we need an api to handle downstream import/exports without css-loader trying to process css-modules. This is also true for the new `auto` default. it's not safe to assume any `.module.css` file should be processed by css-loader when there are loaders ahead of it.

Would love if we could come up with a first class API for this!

evilebottnawi:
I don't understand the problem, please clarify

> This is also true for the new auto default. it's not safe to assume any .module.css file should be processed by css-loader when there are loaders ahead of it.

It is breaking change, you can disable this behavior

jquense:
`css-module-loader` converts css to ICSS `:export` and `:import` rules, css-loader would process those without also processing `@value` and classes to create the JS exports. Now it doesn't because the ICSS plugin is only included when `modules` is `true`. 

`css-module-loader` is incompatible with `css-loader` with modules turned on because both loaders would process the file for css modules when it is already done

evilebottnawi:
@jquense Maybe you need `icss: boolean` (`false` by default), if you specify `icss: true`, it is unable only `icssParser` (no css modules plugins), if you set `modules: true` we enable `icss: true` by default. Is this what you are looking for? Should be very easy to implement

jquense:
> if you set modules: true we enable icss: true by default. 

I think that would work, let me get a PR together if that sounds reasonable

evilebottnawi:
@jquense Go ahead, it is allow fix hacky setup for `css-module-loader`

evilebottnawi:
Fixed, now you can add `{ modules: { auto: false }, icss: true }` and all should be work fine, feel free to feelback

jquense:
Realized that the `exportOnlyLocals` being moved makes SSR hard/impossible for this case as well. Even with the icss options we an't turn on just exports without also including the styles anymore

evilebottnawi:
hm, interesting use case

jquense:
Not sure what the best approach would be for options here that doesn't break the new behavior? Maybe making `icss` accept a 'onlyLocals'` value? It duplicates the option, but would work i guess

[...]

PR #1140, opened 2020-07-28
feat: add `icss` option

jquense:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

#1134 more broadly enabling other loaders to handle CSS Modules

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -111,7 +111,7 @@ export default async function loader(content, map, meta) {
   const icssPluginImports = [];
   const icssPluginApi = [];
 
-  if (needUseModulesPlugins) {
+  if (needUseModulesPlugins || options.icss) {
     const icssResolver = this.getResolve({
       mainFields: ['css', 'style', 'main', '...'],
       mainFiles: ['index', '...'],
--- a/src/options.json
+++ b/src/options.json
@@ -118,6 +118,10 @@
         }
       ]
     },
+    "icss": {
+      "description": "Enables/Disables handling the CSS module interoperable import/export format ((https://github.com/webpack-contrib/css-loader#icss)",
+      "type": "boolean"
+    },
     "sourceMap": {
       "description": "Enables/Disables generation of source maps (https://github.com/webpack-contrib/css-loader#sourcemap).",
       "type": "boolean"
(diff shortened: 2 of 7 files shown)


src/options.json
```diff
@@ -118,6 +118,10 @@
         }
       ]
     },
+    "icss": {
+      "description": "Enables/Disables handling the CSS module interoperable import/export format ((https://github.com/webpack-contrib/css-loader#icss)",
+      "type": "boolean"
+    },
```
evilebottnawi, line 124: Put this option to `modules.icss`, because it is related to CSS modules, not for CSS itself

jquense: that would make the other checks for css modules harder since they all check if `modules` is truthy. e.g.

```js
shouldUseModulesPlugins({ modules }) {
   if (typeof modules === 'object && modules) return options.icss 
   return Boolean(modules)
}
```

Also it's semantically odd, `module: { icss: true }` doesn't enable css modules but any other truthy value would

evilebottnawi: Yep, you are right, let's fix CI

src/index.js
```diff
@@ -111,7 +111,7 @@ export default async function loader(content, map, meta) {
   const icssPluginImports = [];
   const icssPluginApi = [];
 
-  if (needUseModulesPlugins) {
+  if (needUseModulesPlugins || options.icss === true) {
```
evilebottnawi, line 114:
````js
if (needUseModulesPlugins || options.icss) {
  // ...
}
```

src/utils.js
```diff
@@ -197,10 +197,12 @@ function getModulesOptions(rawOptions, loaderContext) {
 }
 
 function normalizeOptions(rawOptions, loaderContext) {
+  const modulesOptions = getModulesOptions(rawOptions, loaderContext);
   return {
     url: typeof rawOptions.url === 'undefined' ? true : rawOptions.url,
     import: typeof rawOptions.import === 'undefined' ? true : rawOptions.import,
-    modules: getModulesOptions(rawOptions, loaderContext),
+    modules: modulesOptions,
+    icss: modulesOptions ? true : rawOptions.icss || false,
```
evilebottnawi, line 205:
```
icss: modulesOptions ? true : rawOptions.icss
```

jquense:
thank you!

evilebottnawi:
ETA is today

PR #1149, opened 2020-07-31
feat: add modules.type and deprecate icss option

### LINKED ISSUE #1134: v4 breaks css-module-loader

jquense:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #1134

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,6 +535,7 @@ module.exports = {
             namedExport: true,
             exportLocalsConvention: 'camelCase',
             exportOnlyLocals: false,
+            type: 'full',
           },
         },
       },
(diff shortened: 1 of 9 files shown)


jquense:
The prop name seems less than ideal, maybe `targets`?

test/fixtures/icss/tests-cases/export/source.css
```diff
@@ -1,3 +1,7 @@
+.other {
```
jquense, line 1: added to check that the class wasn't getting hashed

evilebottnawi:
Done https://github.com/webpack-contrib/css-loader/pull/1150, with some changes, instead `type`, we use `compileType` to avoid misleading

PR #1150, opened 2020-07-31
feat: add `module.type` option, the `icss` option is deprecated

### LINKED ISSUE #1134: v4 breaks css-module-loader

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

fixes #1134

### Breaking Changes

No

### Additional Info

The `icss` option is deprecated


jquense:
awesome, as always thanks for all the hard work 👍

soulofmischief:
@evilebottnawi using `css-loader` without any options now swallows `:export` directives for me. It seems I can specify `'icss'` but then I would lose full module support. Not even sure if that would fix it anyway. I can open a bug report if need be.f

evilebottnawi:
@soulofmischief open an new issue with reproducible test repo

Issue #1161, opened 2020-08-12
Keep :export also for non-CSS-modules


kroko:
- Operating System: nix
- Node Version: v12.17.0
- NPM Version: 6.14.5
- webpack Version: 4.44.1
- css-loader Version: 4.2.1

### Expected Behavior / Situation

Previously my setup was as follows

```javascript
   // --------
    // SCSS GENERAL
    {
      test: /\.(scss)$/,
      exclude: /\.module\.(scss)$/,
      use: [
        // ...
        {
          loader: 'css-loader',
          options: {
            importLoaders: 3,
            sourceMap: true
          }
        },
        // ...
      ]
    },
    // --------
    // SCSS MODULES
    {
      test: /\.module\.(scss)$/,
      use: [
        // ...
        {
          loader: 'css-loader',
          options: {
            importLoaders: 3,
            sourceMap: true,
            modules: {
              mode: 'local',
              localIdentName: (development) ? '[name]__[local]--[hash:base64:5]' : '[hash:base64:16]'
            },
            localsConvention: 'asIs',
            onlyLocals: false
          }
        },
        // ...
      ]
    },
```

and having `:export` directive in non-module scss file worked

_variables.scss_

```scss
:export {
  canvasTouchPointColor: $colorSelectionText;
  canvasBackgroundColor: $colorHeaderBackground;
}
```

_somejavascript.js_

```javascript
import styles from 'shared/scss/variables.scss';
console.log('styles.canvasTouchPointColor', styles.canvasTouchPointColor);
console.log('styles.canvasBackgroundColor', styles.canvasBackgroundColor);
```

### Actual Behavior / Situation

Now, using the new configuration syntax

```javascript
    // --------
    // SCSS GENERAL
    {
      test: /\.(scss)$/,
      exclude: /\.module\.(scss)$/,
      use: [
        // ...
        {
          loader: 'css-loader',
          options: {
            importLoaders: 3,
            sourceMap: true,
            modules: {auto: false} // unnecessary, but let us be verbose 
          }
        },
        // ...
      ]
    },
    // --------
    // SCSS MODULES
    {
      test: /\.module\.(scss)$/,
      use: [
        // ...
        {
          loader: 'css-loader',
          options: {
            importLoaders: 3,
            sourceMap: true,
            modules: {
              mode: 'local',
              localIdentName: (development) ? '[name]__[local]--[hash:base64:5]' : '[hash:base64:16]',
              exportLocalsConvention: 'asIs',
[...]

PR #1162, opened 2020-08-12
examples of icss as well as syncing variables

kroko:
see https://github.com/webpack-contrib/css-loader/issues/1161

added two examples of 
* only icss and no modules
* variables in *only icss* mode and syncing it with both modules as well as JavaScript


kroko:
Deleted this PR and added new https://github.com/webpack-contrib/css-loader/pull/1163 as trying to pass linting created unnecessary many commits.

PR #1163, opened 2020-08-12
docs: example of icss only and mixed

kroko:
### Motivation / Use-Case

See https://github.com/webpack-contrib/css-loader/issues/1161

added two examples for
* only icss, no modules
* both only icss, no modules and modules, sharing variables with both modules as well as JavaScript

### DIFF
--- a/README.md
+++ b/README.md
@@ -1243,68 +1243,12 @@ module.exports = {
 };
 ```
 
-### Using `Interoperable CSS` features only
+### Separating `Interoperable CSS`-only and `CSS Module` features
 
-The following setup is an example of allowing `Interoperable CSS` features such as `:import` and `:export` without using modules by setting `compileType` option.
+The following setup is an example of allowing `Interoperable CSS` features only (such as `:import` and `:export`) without using further `CSS Module` functionality by setting `compileType` option for all files that do not match `*.module.scss` naming convention. This is for reference as having `ICSS` features applied to all files was default `css-loader` behavior before v4.  
+Meanwhile all files matching `*.module.scss` are treated as `CSS Modules` in this example.
 
-**webpack.config.js**
-
-```js
-module.exports = {
-  module: {
-    rules: [
-      // ...
-      // --------
-      // SCSS
-      {
-        test: /\.scss$/,
-        exclude: /\.module\.scss$/,
-        use: [
-          {
-            loader: 'style-loader'
-          },
-          {
-            loader: 'css-loader',
-            options: {
-              importLoaders: 1,
-              modules: {
-                compileType: 'icss'
-              }
-            }
-          },
-          {
-            loader: 'sass-loader'
-          },
-        ],
-      },
-      // --------
-      // ...
-  },
-};
-```
-
-Using SCSS variables in JavaScript
-
-**variables.scss**
-
-```scss
-:export {
-  colorBackgroundCanvas: red;
-}
-```
-
-**app.js**
-
-```js
-import svars from 'variables.scss';
-ctx.fillStyle = `${svars.colorBackgroundCanvas}`;
-```
-
-### Using SCSS variables for both CSS modules and JavaScript if variables are defined in non-module files
-
-The following setup is an example of allowing `Interoperable CSS` features such as `:import` and `:export` without using modules (by setting `compileType` option) together with synchronizing variable values between CSS and Javascript.
-
-Assume case where one wants some canvas drawing to be the same color (set by color name) as HTML background (set by class name).
+An example case is assumed where a project requires canvas drawing variables to be synchronized with CSS - canvas drawing uses the same color (set by color name in JavaScript) as HTML background (set by class name in CSS).
 
 **webpack.config.js**
 
@@ -1314,7 +1258,7 @@ module.exports = {
     rules: [
       // ...
       // --------
-      // SCSS GENERAL
+      // SCSS ALL EXCEPT MODULES
       {
         test: /\.scss$/,
         exclude: /\.module\.scss$/,
@@ -1366,6 +1310,8 @@ module.exports = {
 
 **variables.scss**
 
+File treated as `ICSS`-only.
+
 ```scss
 $colorBackground: red;
 :export {
@@ -1374,6 +1320,8 @@ $colorBackground: red;
 ```
 
 **Component.module.scss**
+
+File treated as `CSS Module`.
 
 ```scss
 @import 'variables.scss';
@@ -1384,16 +1332,18 @@ $colorBackground: red;
 
 **Component.jsx**
 
+Using both `CSS Module` functionality as well as SCSS variables directly in JavaScript.
+
 ```jsx
 import svars from 'variables.scss';
 import styles from 'Component.module.scss';
 
-// render DOM with CSS modules class name
+// Render DOM with CSS modules class name
 // <div className={styles.componentClass}>
 //   <canvas ref={mountsCanvas}/>
 // </div>
 
-// somewhere in JavaScript canvas drawing code use the variable directly
+// Somewhere in JavaScript canvas drawing code use the variable directly
 // const ctx = mountsCanvas.current.getContext('2d',{alpha: false});
 ctx.fillStyle = `${svars.colorBackgroundCanvas}`;
 ```
--- a/README.md
+++ b/README.md
@@ -1261,8 +1261,7 @@ module.exports = {
         exclude: /\.module\.scss$/,
         use: [
           {
-            loader: 'style-loader',
-            options: {}
+            loader: 'style-loader'
           },
           {
             loader: 'css-loader',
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1243,6 +1243,217 @@ module.exports = {
...
+              sourceMap: true
+            }
+          },
+          {
+            loader: 'sass-loader',
+            options: {
+              sourceMap: true
+            }
+          },
+        ],
```
evilebottnawi, line 1415: Remove from  there `resolve-url-loader` and `postcss-loader`, it is unnecessary

README.md
```diff
@@ -1243,6 +1243,217 @@ module.exports = {
...
+              importLoaders: 3,
+              sourceMap: true,
+              modules: {
+                compileType: 'module',
+                mode: 'local',
+                exportGlobals: false,
+                namedExport: false,
+                exportLocalsConvention: 'asIs',
+                exportOnlyLocals: false
+              }
```
evilebottnawi, line 1394: No need extra options in examples

README.md
```diff
@@ -1243,6 +1243,217 @@ module.exports = {
...
+      // ...
+      // --------
+      // SCSS GENERAL
+      {
+        test: /\.scss$/,
+        exclude: /\.module\.scss$/,
+        use: [
+          {
+            loader: 'style-loader',
+            options: {}
```
evilebottnawi, line 1341: No need this

README.md
```diff
@@ -1243,6 +1243,217 @@ module.exports = {
...
+            options: {
+              sourceMap: true
+            }
+          },
+        ],
+      },
+      // --------
+      // ...
+  },
+};
```
evilebottnawi, line 1420: You can simplify this config, just use `auto: /regexp/`

README.md
kroko: done

README.md
kroko: done

README.md
kroko: removed options key, but kept scoped in object for consistency or `use` array members

README.md
```diff
@@ -1243,6 +1243,185 @@ module.exports = {
...
+            options: {
+              sourceMap: true
+            }
+          },
+          {
+            loader: 'resolve-url-loader',
+            options: {
+              sourceMap: true
+            }
+          },
```
evilebottnawi, line 1287: Let's remove unnecessary loaders from here too, example should be simple with only necessary stuff and description

kroko: see the latest commit, it still has `sourceMap` for loaders. can remove `sourceMap` from all of them if necessary.

kroko:
@evilebottnawi did simplify. anything else?

README.md
evilebottnawi: Yes, let's remove `sourceMap`, because it is not related to this

evilebottnawi: `resolve-url-loader` is also unnecessary for this stuff

kroko: will do, give me a sec (in real life scenario source maps are kind of necessary, as some of the common loaders that are used in the css related chains break (postcss plugins and what not) 😉 , at least it was so for a very long time, haven't checked recent "default behaviour" for them)

kroko:
in last pull also fixed `importLoaders` count that was left incorrect after removing unnecessary loaders

README.md
kroko: done

README.md
kroko: hmm?

what we need is that for
- for !`*.module.scss` the files are treated as `compileType: 'icss'` 
- for `*.module.scss` the files are treated as `compileType: 'module'`

by using `auto: "enable only if *.module.scss"` the case would be that
- for !`*.module.scss` the files are treated as neither ICSS nor modules (FAIL - `:export` cannot be used in SCSS)
- for `*.module.scss` the files are treated as `compileType: 'module'` (OK - modules can be used)

therefore we need two `test`s in the loader chain
- for !`*.module.scss` the files are treated as `compileType: 'icss'` (OK - `:export` can be used in SCSS)
- for `*.module.scss` the files are treated as `compileType: 'module'` (OK - modules can be used)

evilebottnawi: Just interesting why you use `module` for icss? Can we improve it and enable `icss` for `file.icss.css`, what do you think?

kroko: I'll try to explain my and my collegues' use case.
There are two types of SCSS files.
Consider standard component/atomized project structure, i.e.,  React.

1)

For each React Container/Component.

```txt
└── Components
    ├── ComponentA
    │   ├── ComponentA.jsx
    │   ├── ComponentA.module.scss
    │   └── index.js
    └── ComponentB
        ├── ComponentB.jsx
        ├── ComponentB.module.scss
        └── index.js
```

These SCSSes should be true CSS modules because of scoping 😉 One can use same "logical" class naming across components (say each component has root container having class "container"), but when they are compiled in a product, they do not clash, I mean you know it already.

2)

"Standartized" global SCSS files that are imported at "root" and deal with
- mixins
- functions
- variables
- bootstraping that is outside the scope of normalize.css
- typography
- styling stuff (including class names) that is outside React root component (and thus has "global" class names)

Take as example... the example provided 😉 
The project has "base variables" such as typefaces, colours, sizing, breakpoints etc. React Containers / Components SCSSes relate to these variables, import them. However, those SCSS variables are also sometimes needed directly in JavaScript - a real life situation is canvas drawings. So for "global SCSS" we need ICSS functionality of being able to use `:export`. Sure, these could be "converted" to modules, but then one has to use `:global`. There are multiple reasons why it would be suboptimal for our case, therefore we _always_ have used two `test`s, just as in the example.

Remember, the reason for this PR is actually only because behaviour of `css-loader` has changed in v4. Before that we also used two `test`s one for all SCSSes excluding `*.module.scss` and one for all SCSSes that are strictly `*.module.scss`, only thing was that before to get ICSS functionality of being able to use `:export` was OOB, one did not have to specify `modules` key at all for ICSS functionality. That is, this worked ->


```javascript
module.exports = {
  module: {
    rules: [
      // ...
      // --------
      // SCSS GENERAL
      {
        test: /\.scss$/,
        exclude: /\.module\.scss$/,
        use: [
          {
            loader: 'style-loader'
          },
          {
            loader: 'css-loader',
            options: {
              importLoaders: 1
            }
          },
          {
            loader: 'sass-loader'
          },
        ],
      },
      // --------
      // SCSS MODULES
      {
        test: /\.module\.scss$/,
        use: [
          {
            loader: 'style-loader'
          },
          {
            loader: 'css-loader',
            options: {
              importLoaders: 1,
              modules: {
                compileType: 'module'
              }
            }
          },
          {
            loader: 'sass-loader'
          },
        ],
      },
      // --------
      // ...
  },
};
```

After upgrading to v4 our _canvases_ broke. Thus I was submitting OP @ https://github.com/webpack-contrib/css-loader/issues/1161 and now it turns out that one has to explicitly specify `modules` key with `compileType: 'icss'`. My example kind addresses this issue.

===

A)

I do think that new naming convention is a logical idea. IMHO, if you implemented that 
- files, that satisfy `/\.module\.(css|scss|less)$/` are automatically set to `compileType: 'module'`
- files, that satisfy `/\.icss\.(css|scss|less)$/` are automatically set to `compileType: 'icss'`
would not hurt nearly anybody, it just needs to be documented in bold 😄 

B)

Other approach would be to set up v4 that by default ICSS is on, just as it was before V4 (thus, if `modules` key is omitted in webpack config, then `css-loader` behaves as if it was

```javascript
modules: {
  compileType: 'icss'
}
```

I cannot thing of any bad side effects if it was so.

That does not change the fact that people should be free to choose filenames. My proposed addition to docs show how to deal with case where `filename.scss` should be treated with ICSS sugar (as it was OOB before V4) and `filename.modules.scss` with all the module stuff.

evilebottnawi: `B` has perf issues, a lot of developers don't use modules or icss, so parsing/anylize/etc cost build time, we should avoid it

kroko: Then you can go with A as a new feature.  

For those who do not comply with the _new naming scheme_, my example still applies.

If you are not going to merge this in, then I can close the PR. Or if you are willing to merge this while also implementing A I can edit the docs

> The following setup is an example of allowing `Interoperable CSS` features only (such as `:import` and `:export`) without using further `CSS Module` functionality by setting `compileType` option for files that filename does not match `/\.icss\.\w+$/i.test(filename)` regular expression. This is for reference as having ICSS features applied to all files was default `css-loader` behaviour before v4.

or smth like that

evilebottnawi: I think we should merge it, and when we implement `auto` for `icss` we will fix docs

kroko: give me a sec then, i'll change some bits to explain why the example exists more clearly (default css-loader behaviour before v4).

kroko: @evilebottnawi done. I also _merged_ the two examples in one.

evilebottnawi:
Thanks!

PR #1181, opened 2020-08-25
feat: add one-letter-css plugin

denisx:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

resolve issue #1028

For efficient gzip/br compression, plugin combine css hash via one symbol name,
as a classname position at file, with filepath hash:base64:8, to have strong sequences.


src/plugins/hash-len-suggest.js
```diff
@@ -0,0 +1,87 @@
...
+    return matchLen;
+  }
+
+  run() {
+    const { instance, selectedHashLen } = this;
+    const matchLen = HashLenSuggest.collectHashLen(instance.getStat());
+
+    console.log();
+    console.log('Suggest Minify Plugin');
+    console.log('Matched length (len: number):', matchLen);
```
evilebottnawi, line 63: We have `logger` for webpack, avoid using `console.log` https://webpack.js.org/api/logging/

src/plugins/hash-len-suggest.js
```diff
@@ -0,0 +1,87 @@
...
+          matchLen[len] = matchLen[len] || 0;
+          matchLen[len] += 1;
+        } else {
+          base[len][hash] = 1;
+        }
+      }
+    });
+
+    return matchLen;
+  }
```
evilebottnawi, line 55: We should use weak cache here, otherwise is was memory leak in wathc mode

src/plugins/hash-len-suggest.js
```diff
@@ -0,0 +1,87 @@
...
+          }).`
+        );
+      }
+
+      console.log();
+    }
+  }
+}
+
+module.exports = HashLenSuggest;
```
evilebottnawi, line 87: I think we can avoid using plugin, we have the identifier of module, so we can generate names based on this

src/plugins/one-letter-css.js
```diff
@@ -0,0 +1,92 @@
...
+
+const loaderUtils = require('loader-utils');
+
+export default class OneLetterCss {
+  constructor() {
+    // Save char symbol start positions
+    this.a = 'a'.charCodeAt(0);
+    this.A = 'A'.charCodeAt(0);
+    // file hashes cache
+    this.files = {};
```
evilebottnawi, line 43: Here potential memory leak in watch mode, if will have a lot of renamed files, the object will grow

denisx: fixed at last call - getStat()

src/plugins/hash-len-suggest.js
denisx: fix

src/plugins/hash-len-suggest.js
denisx: need help with it

src/plugins/hash-len-suggest.js
denisx: we use file path to make long cash. and make shortest hash as we can (but need check for collisions)

src/plugins/hash-len-suggest.js
denisx: @alexander-akait help plz

alexander-akait (commented):
Maybe we can do it using pattern like `localIdentName: `[single-letter]` and keep login there, it will allow more flexibility

denisx:
@MrHBS it is allready at master https://github.com/webpack-contrib/css-loader/commit/ca4abce74d01bc13959e30a4dbbd1ca3a0516d72

tnx to @subzey 

use [v6.6.0](https://github.com/webpack-contrib/css-loader/releases/tag/v6.6.0) or later 💪

PR #1193, opened 2020-09-22
feat: add fallback if custom getLocalIdent returns null

tvsbrent:
This PR contains a:

- [ ] **bugfix**
- [X] new **feature**
- [ ] **code refactor**
- [X] **test update**
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

With the release of the 4.x version of the css-loader, the behavior that allowed a `getLocalIdent` function to return a falsy value to fallback to the default `getLocalIdent` function was lost. That original behavior was introduced in [this pull request](https://github.com/webpack-contrib/css-loader/pull/963).

This PR reintroduces that behavior, but instead requires the function to return `null` or `undefined`.

### Breaking Changes

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -848,6 +848,8 @@ Default: `undefined`
 
 Allows to specify a function to generate the classname.
 By default we use built-in function to generate a classname.
+If the custom function returns `null` or `undefined`, we fall back to the
+built-in function to generate the classname.
 
 **webpack.config.js**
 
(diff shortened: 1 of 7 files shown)


src/utils.js
```diff
@@ -283,11 +283,32 @@ function getModulesPlugins(options, loaderContext) {
...
+            {
+              context: localIdentContext,
+              hashPrefix: localIdentHashPrefix,
+              regExp: localIdentRegExp,
+            }
+          );
+
+          // A null/undefined value signals that we should invoke the default
+          // getLocalIdent method.
+          if (localIdent == null && getLocalIdent !== defaultGetLocalIdent) {
```
evilebottnawi, line 299: Why we need `getLocalIdent !== defaultGetLocalIdent`?

tvsbrent: In the situation where there wasn't a custom `getLocalIdent` function defined and the default function returned `null` or `undefined` for some reason, I didn't want to run the default again. It wouldn't hurt anything to do so, but it also wouldn't help anything either! ;)

evilebottnawi: I think it is bad logic, we don't need `getLocalIdent !== defaultGetLocalIdent`, you can't use `defaultGetLocalIdent` in own code

tvsbrent: Well, at [line 140](https://github.com/tvsbrent/css-loader/blob/96ba438dc19ab833966a0aad46fc00188e49e14d/src/utils.js#L140) in the `utils.js`, the `getLocalIdent` property defaults to `defaultGetLocalIdent` function. So if a user doesn't provide a function, they are implicitly using the default function. The intent for the `!== defaultGetLocalIdent` check is to handle the situation where the user didn't provide a `getLocalIdent` function and `defaultGetLocalIdent` returned a `null` or `undefined`.

That may never happen, as there may already be safeguards in place to keep that from happening - and even if it did run the same function twice, it probably wouldn't be a big deal - so that check can be removed.

README.md
```diff
@@ -848,6 +848,8 @@ Default: `undefined`
 
 Allows to specify a function to generate the classname.
 By default we use built-in function to generate a classname.
+If the custom function returns `null` or `undefined`, we fall back to the
```
evilebottnawi, line 851: `fallback` :smile:

evilebottnawi (dismissed):

tvsbrent:
Sorry, I had to fix a commit subject to correct the lint error.

evilebottnawi:
Thanks!

PR #1271, opened 2021-03-10
Fix inline loader syntax in import

alexander-akait:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #1270

### Breaking Changes

No

### Additional Info

More tests + update deps

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,14 +5,14 @@
   "requires": true,
   "packages": {
     "": {
-      "version": "5.0.2",
+      "version": "5.1.1",
       "license": "MIT",
       "dependencies": {
         "camelcase": "^6.2.0",
         "cssesc": "^3.0.0",
         "icss-utils": "^5.1.0",
         "loader-utils": "^2.0.0",
-        "postcss": "^8.2.6",
+        "postcss": "^8.2.8",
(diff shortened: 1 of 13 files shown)


jquense:
this is working great thanks! Any ideas for solving the `auto` behavior when the request path differs from the virtual resource path?

alexander-akait:
@jquense Can you clarify what is not working? Based on my tests, all works fine...

jquense:
Cases like the original issue:

`import styles from 'Button.module.css!=!./not-modules.css'`

Or

`import styles from 'Button.module.css!=!base64-loader?data=[base64 string]!./button.js'`

Or just

`import styles from 'Button.module.css!=!base64-loader?data=[base64 string]'`


The first 2 compile but don't return modules, the last one doesn't breaks loudly

alexander-akait:
> The first 2 compile but don't return modules

What do you mean, can you provide example?

jquense:
yes here are two test cases: https://github.com/webpack-contrib/css-loader/pull/1273/files I would expect both cases to be processed as css modules

PR #1273, opened 2021-03-11
test: failing test cases for inline syntax and `modules: auto`

jquense:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### DIFF
--- a/test/__snapshots__/loader.test.js.snap
+++ b/test/__snapshots__/loader.test.js.snap
@@ -675,7 +675,7 @@ Array [
     "",
   ],
   Array [
-    "plain.scss!=!../../src/index.js?[ident]!./index-loader-syntax-sass.css",
+    "button.modules.css!=!./index-loader-syntax-sass.css",
     ".baz {
   width: 5px;
 }",
@@ -732,6 +732,20 @@ Array [
   Array [
     "./index-loader-syntax.css",
     ".a {
+  color: red;
+}",
+    "",
+  ],
+  Array [
+    "button.modules.css!=!./index-loader-syntax-sass.css",
+    ".baz {
+  width: 5px;
+}",
+    "",
+  ],
+  Array [
+    "button.module.scss!=!./base64-loader/index.js?[ident]!./simple.js",
+    ".foo {
   color: red;
 }",
     "",
--- a/test/fixtures/base64-loader/index.js
+++ b/test/fixtures/base64-loader/index.js
@@ -1,0 +1,4 @@
+module.exports = function loader(content) {
+    console.log(Buffer.from(this.query.slice(1), 'base64').toString('ascii'))
+    return Buffer.from(this.query.slice(1), 'base64').toString('ascii');
+};
--- a/test/fixtures/index-loader-syntax.js
+++ b/test/fixtures/index-loader-syntax.js
@@ -1,5 +1,11 @@
 import css from './index-loader-syntax.css';
 
-__export__ = css;
+import styles from 'button.modules.css!=!./index-loader-syntax-sass.css'
+
+// the base64 decodes to ".foo { color: red; }"
+import styles2 from './button.module.scss!=!./base64-loader?LmZvbyB7IGNvbG9yOiByZWQ7IH0=!./simple.js'
+
+
+__export__ = [...css, ...styles, ...styles2];
 
 export default css;
--- a/test/loader.test.js
+++ b/test/loader.test.js
@@ -553,6 +553,7 @@ describe("loader", () => {
     expect(getModuleSource("./index-loader-syntax.css", stats)).toMatchSnapshot(
       "module"
     );
+
     expect(getExecutedCode("main.bundle.js", compiler, stats)).toMatchSnapshot(
       "result"
     );


test/__snapshots__/loader.test.js.snap
```diff
@@ -733,6 +733,20 @@ Array [
...
     ".a {
   color: red;
+}",
+    "",
+  ],
+  Array [
+    "button.modules.css!=!./index-loader-syntax-sass.css",
+    ".baz {
+  width: 5px;
+}",
```
jquense, line 743: note they aren't hashed

test/__snapshots__/loader.test.js.snap
```diff
@@ -733,6 +733,20 @@ Array [
...
+    "",
+  ],
+  Array [
+    "button.module.scss!=!./base64-loader/index.js?[ident]!./simple.js",
+    ".foo {
+  color: red;
```
jquense, line 749: ditto

test/__snapshots__/loader.test.js.snap
alexander-akait: Why do not use `index-loader-syntax-sass.modules.css`? Because request here is `index-loader-syntax-sass.css`

jquense: that wouldn't illustrate the problem. css-loader only looks at the resourcePath, not the matchResource. The whole goal of using this syntax is as an alternative to virtual files. **ideally** i'd want this to work in the case where there is no `resourcePath`

jquense: e.g. having this `button.module.scss!=!./base64-loader/index.js?[ident]` work, at least from what Tobias suggested in the other issue. adding a resource to the end is just a hack to get something working

alexander-akait: I think we can't do it, because `button.module.scss` can be any format...

jquense: that's true of any file you point a loader at...it's confusing to me that 'button.module.scss' is used to determine what loaders apply to this resource, but it's not passed to the loaders themselves.

like the goal of the `!=!` syntax seems be for *telling* you what format the file is. 

again the goal would be for `button.module.scss!=!./base64-loader/index.js?[ident]` to work *as is* with css-loader. This syntax _does_ work in webpack, it treats the above as a `.module.scss` file, the problem is css-loader doesn't (because it has no info)

alexander-akait: Please describe what you want to solve

jquense: i've describe what i want to solve many times now in this issue: https://github.com/webpack/webpack/issues/11074

please let me know if there is something else i can be clearer on

jquense: in a nutshell I want to turn: 

```js
const styles = css`
  .foo {
    color: red
  }
`
```
into 

```js
styles = require('./generated-file-that-doesn't-exist-on-disk.module.css')
```

I was using virtual modules to do this, @sokra said, that was a bad idea and i should turn it into:

```js
styles = require('.styles.module.css!=!content-from-query-loader?"foo { color: red;}"!')
```

I am now trying to make that work with css-loader

alexander-akait: Why do not use `styles = require('styles.module.css!=!content-from-query-loader?"foo { color: red;}"!./original-file.modules.js')`?

So we got `original` file when move content to you loader where you do transformation, for example - extract CSS, when we move it to `styles.module.css` and apply CSS loader?

jquense: the JS file is never named `.modules.js`, since that's not a common naming pattern. I can't really tell users or my team to name their React component files `.module.js`, and it's  not possible if the original file is a .svelte, or .vue file or other mixed style format

alexander-akait: hm, sounds reasonable, give me time to look at this

jquense: It might be the issue is that webpack needs to expose matchResource to loaders, so they can decide which to use

jquense: uncovered another subtle problem with using the original file as the resource with css modules:

multiple virtual files from one original gives css-loader different content with the same resource. may affect caching but also affects how class names are hashed,

```js
// becomes button.module.css!=!./my-loader/!.button.js
const button = css`
    .foo {
     color: blue
   }
`
// becomes button2.module.css!=!./my-loader/!.button.js
const button2 = css`
   .foo {
     color: red
   }
`
```

The above, would produce the same hash for each `.foo` (since the same file is used to generate the local name) breaking the encapsulation of the style

alexander-akait: due this https://github.com/webpack-contrib/css-loader/blob/master/src/utils.js#L73

alexander-akait: What about `button2.module.css?modules=true!=!./my-loader/!.button.js`?

alexander-akait: You can implement `modules` on own side

jquense: Like reimplement it? I could do that, but it's a lot of work. The hope/goal was to use the users existing config. If we need to reimplement css-loader, I can avoid all of this and write the whole thing as a plugin probably but that's more config and more duplicated work. It also doesn't solve the problem for anyone else trying to use virtual modules in this way, like svelte

jquense: the real hard part though of that plain is when there are things in front of css-loader:

```js
button2.module.css!=!./my-loader?modules=true!.button.js
```
resolves fully to:
```
css-loader!sass-loader!my-loader?modules!./button.js
```

`my-loader` can't do modules at it's spot because Sass still needs to run. my loader also can't enable modules on the lower css-loader because it would change the option for all files. The only thing it can is something like: https://github.com/webpack/webpack/issues/10350 which i say works great, but have since realized *does not work* well at scale for a few, hard to describe, reasons that causes styles to be included multiple times and out of order, due to module `id`'s being (correctly) different between the plain require and compiled requires

PR #1274, opened 2021-03-15
fix: the `auto` option for inline module syntax

alexander-akait:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #1273

### Breaking Changes

No

### Additional Info

/cc @jquense fixed, please look

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -69,11 +69,17 @@ function defaultGetLocalIdent(
   options
 ) {
   const { context, hashPrefix } = options;
+  // eslint-disable-next-line no-underscore-dangle
+  const { matchResource } = loaderContext._module;
   const { resourcePath } = loaderContext;
   const request = normalizePath(path.relative(context, resourcePath));
 
   // eslint-disable-next-line no-param-reassign
-  options.content = `${hashPrefix + request}\x00${localName}`;
+  options.content = `${
+    hashPrefix +
+    (typeof matchResource !== "undefined" ? `${matchResource}\x00` : "") +
+    request
+  }\x00${localName}`;
 
   return interpolateName(loaderContext, localIdentName, options);
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -68,28 +68,22 @@ function defaultGetLocalIdent(
   localName,
   options
 ) {
-  let matchResourceRequest;
+  let relativeMatchResource = "";
 
-  // // eslint-disable-next-line no-underscore-dangle
-  // if (loaderContext._module.matchResource) {
-  //   matchResourceRequest = normalizePath(
-  //     // eslint-disable-next-line no-underscore-dangle
-  //     path.relative(options.context, loaderContext._module.matchResource)
-  //   );
-  // }
+  // eslint-disable-next-line no-underscore-dangle
+  if (loaderContext._module.matchResource) {
+    relativeMatchResource = `${normalizePath(
+      // eslint-disable-next-line no-underscore-dangle
+      path.relative(options.context, loaderContext._module.matchResource)
+    )}\x00`;
+  }
 
-  const request = normalizePath(
+  const relativeResourcePath = normalizePath(
     path.relative(options.context, loaderContext.resourcePath)
   );
 
   // eslint-disable-next-line no-param-reassign
-  options.content = `${
-    options.hashPrefix +
-    (typeof matchResourceRequest !== "undefined"
-      ? `${matchResourceRequest}\x00`
-      : "") +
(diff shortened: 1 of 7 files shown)


test/__snapshots__/loader.test.js.snap
```diff
@@ -733,6 +733,20 @@ Array [
     "./index-loader-syntax.css",
     ".a {
   color: red;
+}",
+    "",
+  ],
+  Array [
+    "button.modules.css!=!./index-loader-syntax-sass.css",
+    ".iuzmhXWrT3_Z-zKS-rMGM {
```
alexander-akait, line 741: Now they are CSS moduled

src/utils.js
```diff
@@ -126,7 +126,10 @@ const moduleRegExp = /\.module(s)?\.\w+$/i;
 const icssRegExp = /\.icss\.\w+$/i;
 
 function getModulesOptions(rawOptions, loaderContext) {
-  const { resourcePath } = loaderContext;
+  const resourcePath =
+    // eslint-disable-next-line no-underscore-dangle
+    loaderContext._module.matchResource || loaderContext.resourcePath;
```
jquense, line 131: woohoo! should this also be the logic for local ident generation?

alexander-akait: Yep, you are right, otherwise we will get the same class names.

After fix local ident generation should we merge it and release? All is fixed?

jquense: i think so, seems like this covers it all. hard to know for sure until i try it out on a code base. Thanks again for helping me out here

alexander-akait: Yep, here interesting case, when we have: 
```
import three from './button.module.scss!=!./base64-loader?LmZvbyB7IGNvbG9yOiByZWQ7IH0=!./simple.js';
import four from './other.module.scss!=!./base64-loader?LmZvbyB7IGNvbG9yOiByZWQ7IH0=!./simple.js';
```

We run the loader only once, why? Because ES modules executed only once. So I don't think here bug. But we can solve it using:
```
// i.e. adding `?` with different values
import three from './button.module.scss!=!./base64-loader?LmZvbyB7IGNvbG9yOiByZWQ7IH0=!./simple.js?foo=bar';
import four from './other.module.scss!=!./base64-loader?LmZvbyB7IGNvbG9yOiByZWQ7IH0=!./simple.js?foo=baz';
```

jquense: ah interesting, so the query param acts like a cache buster. That solution seems fine to me

alexander-akait: Yep, you can generate it and please avoid using `Math.random()`, because cache will be broken

jquense: yeah in my case it'd be easy to use a stable identifier here. This wouldn't solve the className hash generation issue tho right? I'm guessing you discovered this righting a test for that?

alexander-akait: Solved, please check code

test/__snapshots__/loader.test.js.snap
```diff
@@ -733,6 +733,27 @@ Array [
...
+    "button.module.scss!=!./base64-loader/index.js?[ident]!./simple.js?foo=bar",
+    "._1Z1CpM4SYZX9egsUO4gNTz {
+  color: red;
+}",
+    "",
+  ],
+  Array [
+    "other.module.scss!=!./base64-loader/index.js?[ident]!./simple.js?foo=baz",
+    "._1KqQu-H04lA8NLg4-NKnDf {
+  color: red;
```
alexander-akait, line 756: You can see here: hash for classes are different

alexander-akait: But note: you should use different `matchResource`, otherwise hash will be same

jquense: perfect!

jquense:
LGTM

src/utils.js
jquense: arg, i'm trying this out now and `_module.matchResource` is null locally. Going to debug a bit to see why will report back

alexander-akait: If you don't have `matchResource` in request, it will be nullish

jquense: yeah this is for requests with it in it. For some reason i am seeing the module get created twice, one with the matchResource and one without, the one without is what makes it to the css-loader. It maybe has to do with ESM, still trying to figure it out

Issue #1236, opened 2020-12-02
Support Asyncronous functions for config.url and config.import


DavidArchibald:
I don't believe these details should effect replication but I've included them just in case:

- Operating System: Windows
- Node Version: v10.19.0
- NPM Version: 6.14.0
- webpack Version: 4
- css-loader Version: 5.0.0

### Expected Behavior / Situation

I expected to be able to use asynchronous code for filtering URLs and imports, for example:

```js
{
  loader: "css-loader",
  options: {
    url: async (url, contextPath) => ...
  }
}
``` 

The specific use case I'm running into is some Webpack utilities that rely on callbacks like `compiler.inputFileSystem.stat` which can easily be turned into a Promise and then awaited. I'm using complex file system searching due to a custom resolver I'm developing and thusly its interface with css-loader. Disregarding the specific and looking into the general, I don't see this as an adverse design choice to make since a lot of code can be asynchronous for a lot of reasons. It's probably this way because filtering was assumed to be relatively simple and blocking.

### Actual Behavior / Situation

The asynchronous functions aren't awaited and so behave the same as `url: true` because any promise is truthy.

### Modification Proposal

It turns out the loader is already an asynchronous function so this isn't really much of a change. Here's where I believe the filtering is done for `options.import`:

https://github.com/webpack-contrib/css-loader/blob/db3911be1a797abeb3afd5499684b9e610faacaa/src/index.js#L77

Here's the one for `options.url`:

https://github.com/webpack-contrib/css-loader/blob/db3911be1a797abeb3afd5499684b9e610faacaa/src/index.js#L104

I believe to solve this would be as trivial as prepending the function call with `await`, i.e. doing `await getFilter(...)` because in the case of non-promises Node just returns the result as per normal, e.g. `await () => false` and others are safe pieces of code. I looked for the specification backing this but apparently this second hand source [StackOverflow answer](https://stackoverflow.com/a/38510353) may be the remains now that the proposal text goes to a deadlink on the [proposal's Github](https://github.com/tc39/proposal-async-await).

Stylistically these specifics doesn't really matter to me but due to what I believe to be a literally 12 character addition I didn't open a PR directly because style concerns or just conversation around this could easily dwarf the actual contribution size significantly and require being redone entirely. I'd be happy to open a PR if that would be helpful of course.

alexander-akait:
Feel free to send a PR

PR #1277, opened 2021-03-20
Support asyncronous functions in `config.url` and `config.import`.

DavidArchibald:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update**
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Fixes https://github.com/webpack-contrib/css-loader/issues/1236.

### Breaking Changes

None

### Additional Info

My apologies for the issue lingering open. Apparently I must've cleared all my notifications and missed your response. I don't believe anything like the schema needs updating (async functions are instance of Function) so the change is pretty small.

### DIFF
--- a/src/plugins/postcss-import-parser.js
+++ b/src/plugins/postcss-import-parser.js
@@ -157,9 +157,12 @@ const plugin = (options = {}) => {
               media = valueParser.stringify(mediaNodes).trim().toLowerCase();
             }
 
-            if (options.filter && !options.filter(normalizedUrl, media)) {
-              // eslint-disable-next-line no-continue
-              continue;
+            if (options.filter) {
+              const processURL = await options.filter(normalizedUrl, media);
+              if (!processURL) {
+                // eslint-disable-next-line no-continue
+                continue;
+              }
             }
 
             node.remove();
--- a/src/plugins/postcss-import-parser.js
+++ b/src/plugins/postcss-import-parser.js
@@ -157,41 +157,50 @@ const plugin = (options = {}) => {
               media = valueParser.stringify(mediaNodes).trim().toLowerCase();
             }
 
-            if (options.filter) {
-              const processURL = await options.filter(normalizedUrl, media);
-              if (!processURL) {
-                // eslint-disable-next-line no-continue
-                continue;
-              }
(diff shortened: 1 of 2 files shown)


DavidArchibald:
Ah, I thought I had run the validations beforehand but must've missed the lint (it can sometimes be tricky to know what automated checks will be run). Little embarrassing to need two commits for such a small change so I'd expect they should be smushed or combined with --force.

Looks like you have the [no-await-in-loop](https://eslint.org/docs/rules/no-await-in-loop) rule enabled. Would it be appropriate to add a commit that ignores that with `// eslint-disable-next-line no-await-next-line` or should it be refactored inside of the `tasks` promise array below? Promises that are skipped would have to be filtered out of the below loop although that's not complex. I'm asking because it's ultimately stylistic and I don't want to push assuming one way or another. Being so small a lot of routine questions and review can overwhelm the actual work put into the code, haha.

Lint errors for reference:
```
[lint:js      ] /home/runner/work/css-loader/css-loader/src/plugins/postcss-import-parser.js
[lint:js      ]   161:34  error  Unexpected `await` inside a loop  no-await-in-loop
[lint:js      ] 
[lint:js      ] /home/runner/work/css-loader/css-loader/src/plugins/postcss-url-parser.js
[lint:js      ]   264:32  error  Unexpected `await` inside a loop  no-await-in-loop
```

alexander-akait (commented):
`await` inside `for of`, it is bad for performance, we need refactor code and use `Promise.all`

alexander-akait (commented):
Thanks, I will look at this in near future

DavidArchibald:
Sounds good, hope it all adheres now (fail the first commit's message format). As a side note, I struggled with CRLF vs LF differences in the test, I'm assuming because the test snapshots are formatted with Linux/Mac line endings. WSL didn't seem to help in this case either. I struggled with that for a while before deciding to just ignore test failures complaining about `\r\n` vs `\n`. Do you have ideas about how to fix that, perhaps a Jest configuration?

alexander-akait (approved):

PR #1330, opened 2021-07-14
Issue 1307

### LINKED ISSUE #1307: localIdentName setting should only rename classnames for (s)css module files
- Operating System: OSX
- Node Version:  12.18.1
- NPM Version: 6.14.5
- webpack Version: 5.36.2
- css-loader Version: 5.2.4

### Expected Behavior

When setting a `localIdentName` in the `modules` config it should only rename classnames for (s)css modules. 
So files that have .module.(s)css

### Actual Behavior

It also renames classnames  in scss that are passed as webpack entry. So app.scss has some global css and all of the sudden they get hashed. But the react components have the actual classnames and not a reference to imported styles.

### Code

```js
// webpack.config.js

test: /\.s?css$/,
use: [
    {
        loader: require('mini-css-extract-plugin').loader,
    },
    {
        loader: 'css-loader',
        options: {
            sourceMap: true,
            modules: {
                localIdentName:
                    nodeEnv === 'production'
                        ? '[hash:base64]'
                        : '[name]__[local]--[hash:base64:5]',
            },
        }
    },
    {
        loader: 'postcss-loader',
    }
]
```

### How Do We Reproduce?

Create some global css and set this as an entry along with a js entry:

```
    entry: {
            app: [
                path.resolve(__dirname, 'src/scss/app.scss'),
                path.resolve(__dirname, 'src/js/app.js'),
            ],
    },
```

and create a css module which you import in the js file.
With the loader settings above you would expect that only the module css would get renamed.

alexander-akait:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes #1307

### Breaking Changes

Yes, will be described in readme

### Additional Info

No

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -491,8 +491,8 @@ function getValidLocalName(localName, exportLocalsConvention) {
   return camelCase(localName);
 }
 
-const moduleRegExp = /\.module(s)?\.\w+$/i;
-const icssRegExp = /\.icss\.\w+$/i;
+const IS_MODULES = /\.module(s)?\.\w+$/i;
+const IS_ICSS = /\.icss\.\w+$/i;
 
 function getModulesOptions(rawOptions, loaderContext) {
   const resourcePath =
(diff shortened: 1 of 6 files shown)


alexander-akait:
Close in favor https://github.com/webpack-contrib/css-loader/pull/1332, wrong solution

PR #1382, opened 2021-10-01
Feat/hash serialization

subzey:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Current implementation incorrectly uses the hash buffer as an entropy source:

```js
  const localIdentHash = hash
    .digest(hashDigest)
    .slice(0, hashDigestLength)
    .replace(/[/+]/g, "_")
    .replace(/^\d/g, "_");
```

Instead of tossing away unacceptable values they are replaced with other ones. This results a probability "skew" that is most noticeable for a first character. For base64 the alphabet (to pick the pseudorandom value from)

```
A B C D E F G H I J K L M N O P
Q R S T U V W X Y Z a b c d e f
g h i j k l m n o p q r s t u v
w x y z 0 1 2 3 4 5 6 7 8 9 + /
```

becomes

```
A B C D E F G H I J K L M N O P
Q R S T U V W X Y Z a b c d e f
g h i j k l m n o p q r s t u v
w x y z _ _ _ _ _ _ _ _ _ _ _ _
```

As you can see, the underscore has significantly more chances to be picked. It raises the collision probability for the first character from 1.56% (1 / 64) up to 4.79%!

All the subsequent characters are affected, too. They are drawn form the alphabet with two underscores:

```
A B C D E F G H I J K L M N O P
Q R S T U V W X Y Z a b c d e f
g h i j k l m n o p q r s t u v
w x y z 0 1 2 3 4 5 6 7 8 9 _ _
```

*The proposed solution is to:*
- Throw away the unacceptable leading digits instead of trying to replace them.
- Replace `/` → `_` and `+` → `-` (instead of both `/` and `+` → `_`)

As the parts of digest are tossed away, there is a (small) chance it won't be enough to generate the string of the desired length.

To overcome this issue the string is now hashed several times until it's long enough. Each hashing tier has its unique sequential salt. For MD4 / MD5 the max string length generated this way is 68 719 476 736. That should be enough to throw away any amount of unacceptable characters.

_This is essentially a [PBKDF / PBKDF2](https://en.wikipedia.org/wiki/PBKDF2) but with 1 round and indefinite length: the length grows on demand._

Most of the time only one hashing is required so this change should not significantly impact the performance.

### Breaking Changes

:warning:  The `localIdentHash` now may contain dashes (`-`), including leading and trailing ones. It should be okay for classnames.

:warning: The hashing has changed and all the classnames gets another hashes. (The snapshots were updated accordingly). This may come as a surprise for users.

### Additional Info

I've made a synthetic test to show this PR actually improves something:

https://gist.github.com/subzey/3f1a0efd1634bdf0b03ba2a2e744bc3e

Both implementations are called until the first collision. This is repeated 101 times then a median value is picked. This value corresponds to a 50% collision chance.

Current implementation can generate ≈ *20200* hashes until the 50% collision chance.

The proposed implementation can generate ≈ *35600* hashes until the 50% collision chance.

The expected value for true random is ≈ 35440.


alexander-akait:
> warning The localIdentHash now may contain dashes (-), including leading and trailing ones. It should be okay for classnames.

we should avoid it

> warning The hashing has changed and all the classnames gets another hashes. (The snapshots were updated accordingly). This may come as a surprise for users.

it is fine

subzey:
> we should avoid it [the dash]

Okay, that's actually fairly easy to do - just throw away the dashes with the leading digits :D Give me a second

subzey:
@alexander-akait Done in 101989a3afa313e9084cf51ece6f965fb6edef28 (impln) and 817368b9fed1f3ad1d7793acc6f2fc02262e38e5 (snaps)

subzey:
Looks like the [Webpack@5.54.0](https://github.com/webpack/webpack/releases/tag/v5.54.0) introduced a new hash `xxhash64` and that's the reason the hex hash in the filenames doesn't match.

On the other hand, `xxhash64` doesn't accept the `Uint32Array` as input data, I'll fix that shortly in this PR.

subzey:
Updated webpack in package.json to actually match `webpack@latest` in tests. Updated the test to contain a now-default `xxhash64` digest.

Fixed `defaultGetLocalIdent` to work with `xxhash64`. And a new test that it actually works.

alexander-akait:
Thanks

PR #1397, opened 2021-11-03
add syntax option

jazzyclimber:
Allows for custom syntax to be obeyed from postcss-loader and postcss.


alexander-akait:
Sorry, we support only CSS syntax, if you want to use any custom, please use postcss-loader (we reuse cache between them so you should not have perf problems), sorry for delay in review

PR #1410, opened 2022-01-26
feat: do not hash localName if the localName is inlined

subzey:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Proposal: The `localName` should not affect the hash if it is present in the generated id inline. This way the generated identifiers are compressed better with both GZIP and Brotli.

----

Currently the hash is computed from the file relative path and the local identifier name. It's good when the `[local]` is not preset in the template `localIdentName`:

> ✔️ `[hash]`, before the change
`.header` from `one.css` becomes `.ABCDEF`
`.button` from `one.css` becomes `.GHIJKL`
`.button` from `two.css` becomes `.MNOPQR`

But it is redundant if the `[local]` is present:

> 😔 `[local]--[hash]`, before the change
`.header` from `one.css` becomes `.header--ABCDEF`
`.button` from `one.css` becomes `.button--GHIJKL`
`.button` from `two.css` becomes `.button--MNOPQR`

The `.button` from `one.css` and `.button` from `two.css` still have to have different hashes.

But the `.header` from `one.css` and `.button` from the same `one.css` can never collide, they're from the same module! Omitting localName from the hash we get:

> ✔️ `[local]--[hash]`, after the change
`.header` from `one.css` becomes `.header--UVWXYZ`
`.button` from `one.css` becomes `.button--UVWXYZ`
`.button` from `two.css` becomes `.button--FGHIJK`

Now all identifiers from the same module has the same hash. (If and only if the `localName` would be substituted into the resulting generated identifier!) And there are less unique random string runs in the output, making the result more  GZIP friendly.

I've used `test/__snapshots__/modules-option.test.js.snap` for my "gzippability" estimations. This file "before" and "after" has the same "raw" byte length. But the "after" is **almost 8% smaller after being GZIPped**

_We use the similar technique on tradingview.com for a year or so -- using custom `getLocalIdent()` in the webpack settings. But now it's time to reveal one of our optimization secrets 😀_

### Breaking Changes

None

### Additional Info

~~⚠️ The generated identifiers will change compared to the previous versions - only if the `[local]` is used in the template.~~ -- Postponed to the next major release.


alexander-akait:
Let's move this under options to avoid breaking changes, we can set it by default in the next major release

subzey:
Sounds reasonable! How should I better do and name it?

Adding a new setting `localNamesInHash`: `false | true | "auto"` with the default `true`. And then in the next major release the default becomes `"auto"`.

Or should it be something more generic like `hashStrategy`: `"include-local-name" | "omit-local-name" | "auto"`?

alexander-akait:
I think `options.modules.hashStrategy` sounds good

test/__snapshots__/modules-option.test.js.snap
```diff
@@ -3397,6 +3397,766 @@ Array [
...
+// Exports
+___CSS_LOADER_EXPORT___.locals = {
+	\\"123\\": \\"iHMJbI42\\",
```
subzey, line 3791: On a second thought, it's hard to imagine a use-case when `'omit-local-name'` could be used:

- If `[local]` is used, it works just like `'auto'`.
- If `[local]` is not there, it screws everything up.

Should we keep it for completeness and probably #1181 like features?

Or is it a bad idea too keep senseless options and it should better be removed?

alexander-akait: Let's avoid senseless options, we can add them late, if we will need

subzey: Done, in 976ff89756b31035f19e8a441ee55378f255ba76

alexander-akait:
@subzey Looks good, let's regenerate lock file due security problem in CI :+1:

subzey:
EPERM on Windows? I hope we could just restart the test and have it fixed

alexander-akait:
I fixed the problem

README.md
```diff
@@ -872,6 +872,36 @@ module.exports = {
...
 
+##### `hashStrategy`
+
+Type: `'use-local-name' | 'auto'`
+Default: `'use-local-name'`
+
+Should local name be used when computing the hash.
+
+- `'use-local-name'` Each identifier in a module gets its own hash digest.
+- `'auto'` Identifiers from the same module shares the same hash digest if it's possible. Use this value to optimize the output for better GZIP or Brotli compression.
```
alexander-akait, line 883: After review, I think `'auto'` is not good name, `'auto'` sounds like it should be default value, yes, maybe this should be the default (as I written above), but let's found better name, I think we can do it more understandable, I suggest: `resource-path` and `resource-path-and-local-name`, the developer can immediately understand what is used from the name, instead of reading the documentation

subzey: There's a chance a developer would set a `resource-path`, but the hashing would work as `resource-path-and-local-name` because of the automagic circuit breaker. That may lead to confusion, especially when copypasting the config partially into StackOverflow or GH issues.

Maybe `minimal-subset` or `infer-from-template` or something like that? Something that implies the automatic nature of the setting

_I'm not a big fan of bikeshedding and just will use your suggestion if we don't come up with a decision after a short while_

alexander-akait: Let's use `minimal-subset` :+1:

subzey: Renamed the option enum values in 381d470729f54b74657650bb7eea7370ad9f7ad9

alexander-akait:
Thanks

denisx:
tnx for this pr 😍

Issue #1422, opened 2022-02-18
Allow to hook into classname generation after default interpolation logic pass


SkReD:
### Feature Proposal
Two proposal variants:
* Export [`defaultGetLocalIdent`](https://github.com/webpack-contrib/css-loader/blob/master/src/utils.js#L327) 
* Allow to pass [`customInterpolateName`](https://github.com/webpack/loader-utils/blob/9fe238119274581649b78e5a4d9d8e4c76229e91/lib/interpolateName.js#L103) as option to loader

### Feature Use Case
Replace className part which is duplicated across components with some short placeholder:
`[path][name]__[local]` -> `src/blocks/button_link` and replace with `cmp-button_link`

With current options it is not possible without copy pasting code from `defaultGetLocalIdent`.
`localIdentRegExp` is not flexible enought for this case because it not allow to pass replacement string.

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information
  System:
    OS: Windows 10 10.0.19043
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 9.51 GB / 23.86 GB
  Binaries:
    Node: 14.18.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 6.14.15 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 98.0.4758.102
    Edge: Spartan (44.19041.1266.0), Chromium (98.0.1108.55)
    Internet Explorer: 11.0.19041.1202
  Monorepos:
    Yarn Workspaces: 1.22.17
  Packages:
    karma-webpack: ^5.0.0 => 5.0.0
    webpack: ^5.35.0 => 5.57.1
    webpack-cli: ^4.6.0 => 4.7.2
    webpack-merge: ^5.7.3 => 5.7.3

alexander-akait:
Why don't use custom function? It is very exotic

SkReD:
I want reuse default interpolation logic. Dont want to reimplement all escaping things and other stuff

alexander-akait:
Okay, let's do:
- Export [defaultGetLocalIdent](https://github.com/webpack-contrib/css-loader/blob/master/src/utils.js#L327)

alexander-akait:
Fixed

PR #1423, opened 2022-02-21
export defaultGetLocalIdent

SkReD:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
https://github.com/webpack-contrib/css-loader/issues/1422

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1297,4 +1297,5 @@ export {
   camelCase,
   stringifyRequest,
   isDataUrl,
+  defaultGetLocalIdent,
 };
--- a/src/cjs.js
+++ b/src/cjs.js
@@ -1,3 +1,5 @@
 const loader = require("./index");
 
+module.exports.defaultGetLocalIdent = require("./utils").defaultGetLocalIdent;
+
 module.exports = loader.default;


src/utils.js
```diff
@@ -324,7 +324,7 @@ function escapeLocalIdent(localident) {
   );
 }
 
-function defaultGetLocalIdent(
+export function defaultGetLocalIdent(
```
alexander-akait, line 327: We should not export it here, we should reexport it from the entry point, i.e. `index.js`

SkReD: It is done here https://github.com/webpack-contrib/css-loader/pull/1423/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R30

alexander-akait: We should not export it from `utils.js`, we can rename file or move it to another places, just reexport it from `index.js`

SkReD: This is exactly what i did. But re-export in index it should be exported from utils too

SkReD: ready to review. Sorry for the mess

alexander-akait (approved):

PR #1459, opened 2022-06-18
Update README.md

hencca:
It is not true, that to use css-loader webpack 5 is needed.

This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [x] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I was just brushing up on my webpack skills, and noticed this error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,6 @@ The `css-loader` interprets `@import` and `url()` like `import/require()` and wi
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


alexander-akait:
I think we should rewrite it - `To use the latest version of css-loader you need webpack v5`

PR #1470, opened 2022-09-09
ci: add minimum GitHub token permissions for workflows

boahc077:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using https://github.com/step-security/secure-workflows.

The GitHub Actions workflow has a GITHUB_TOKEN with write access to multiple scopes.
Here is an example of the permissions in one of the workflow runs:
https://github.com/webpack-contrib/css-loader/runs/8160728230?check_suite_focus=true#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflow:

- cancel.yml

The following workflow file already has the least privileged token permission set:

- nodejs.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- GitHub recommends defining minimum GITHUB_TOKEN permissions.
  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -1,7 +1,12 @@
 name: Cancel
 on: [push]
+permissions:
+  contents: read
+
 jobs:
   cancel:
+    permissions:
+      actions: write # for styfle/cancel-workflow-action to cancel/stop running workflows
     name: Cancel Previous Runs
     runs-on: ubuntu-latest
     timeout-minutes: 3


snitin315 (dismissed):

boahc077:
@snitin315 thanks so much for reviewing my PR. I looked at the failed lint check. It appears to be a Node linter. Whereas the change I am making is related to GitHub actions which is uses YAML. Is there any way to add exceptions for such failures as it may not applicable for this PR?

snitin315:
lint is failing for the cancel.yml file so it is related. Please run prettier locally to fix lint.

boahc077:
@snitin315 thanks for the pointer. I ran prettier locally to fix cancel.yaml. There was an extra space in my YAML comment. I made the fix and squashed my previous two commits. Please take a look again.

snitin315 (approved):

alexander-akait (approved):

Issue #1384, opened 2021-10-08
SSR: generated class names are not consistent for client and server bundles since v5.1.3


OliverJAsh:
### Bug report

We are using `css-loader` with server-side rendering (SSR), so we run webpack twice: once for the client and once for the server.

For the client, we need to generate CSS files, so we use `mini-css-extract-plugin` in combination with `css-loader`.

For the server, we don't need to generate CSS files, so we use `exportOnlyLocals: true`. We just need the CSS modules to export the same class names as the ones used in the CSS files we generated for the client (above).

### Actual Behavior

Generated class names are not consistent for client and server bundles.

### Expected Behavior

Generated class names are consistent for client and server bundles.

### How Do We Reproduce?

Full reduced test case: https://github.com/OliverJAsh/webpack-css-loader-ident-hash-bug

`package.json`:
```json
{
  "dependencies": {
    "css-loader": "^6.3.0",
    "mini-css-extract-plugin": "^2.4.2",
    "webpack": "^5.58.1",
    "webpack-cli": "^4.9.0"
  }
}
```

`webpack.config.js`:
```js
const MiniCssExtractPlugin = require("mini-css-extract-plugin");

const { MODE } = process.env;

module.exports = {
    plugins: [new MiniCssExtractPlugin()],
    mode: "production",
    entry: "./src/main.js",
    module: {
        rules: [
            {
                test: /\.css$/,
                exclude: /node_modules/,
                use: [
                    ...(MODE === "client" ? [MiniCssExtractPlugin.loader] : []),
                    {
                        loader: "css-loader",
                        options: {
                            modules: {
                                exportOnlyLocals: MODE === "server",
                            },
                        },
                    },
                ],
            },
        ],
    },
};
```

`src/main.js`:
```js
import styles from './main.css'
console.log(styles.container);
```

`src/main.css`:
```css
.container {
    display: block;
}
```

Since [this change which first appeared in v5.1.3][change], `css-loader` generates different class names for the client/server:

```sh-session
$ rm -rf dist && MODE=client webpack && node dist/main.js
UgjPTt4sSFtvFL9kN_LV

$ rm -rf dist && MODE=server webpack && node dist/main.js
BZjOBwj614XLb60af45_
```

It seems this is because `css-loader` now uses `relativeMatchResource` as part of the hash, and in this case, `relativeMatchResource` will be different depending on whether or not we're using `mini-css-extract-plugin`:

[...]

PR #1480, opened 2022-11-10
fix: hash generation with inline syntax

alexander-akait:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

fixes https://github.com/webpack-contrib/css-loader/issues/1384

### Breaking Changes

No

### Additional Info

No


cherepanov:
I see regression on one of our project since version 6.7.2
In combination with style-loader class name hash mismatches with html.

```.close___b9e73c8a63cb4434d320``` vs ```close___defedee1e4d1a20dcfb3```

Really nothing special in config.
Very confused because all apps use the same build config, but only one of them is affected.

Maybe related to SSR

Rollback to 6.7.1 solves issue

alexander-akait:
@cherepanov Can you provide example? Do you mean you classes are overllaping or just changed?

PR #1487, opened 2022-12-13
fix: remove `sourceURL` from emitted CSS

alan-agius4:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

In this case the `sourceURL` is not needed as the entire sourcemap which cotains all the information is being embeeded.

Multiple `sourceURL` is also causing issues when in Chrome devtools.
### Additional Info
See: https://github.com/angular/angular-cli/issues/24414 for more information.


alexander-akait:
Can you create reproducible example using pure webpack, because I can't get two `sourceURL`

```
/*# sourceURL=webpack://./src/app/b/b.component.scss */
/*# sourceURL=webpack://./src/styles.scss */
```

alan-agius4:
@alexander-akait all you need to do is enable sourcemaps in sass-loader. 

Here's an example

https://github.com/webpack-contrib/css-loader/compare/master...alan-agius4:css-loader:repro#diff-b15393677601d8b57053e1e25f5e3e9ef695b570273563e83f2217315bf28cf1R2520-R2522

alexander-akait:
hm, maybe you are right, I looked at code and found we already built-in source map:
```
const data = `sourceMappingURL=data:application/json;charset=utf-8;base64,${base64}`;
const sourceMapping = `/*# ${data} */`;
```

and then duplicate:
```
const sourceURLs = cssMapping.sources.map(
  (source) => `/*# sourceURL=${cssMapping.sourceRoot || ""}${source} */`
);
```

Not sure why we are doing it, I want to look at github history and found why it was implemented

alan-agius4:
Correct, I did look into the git history before doing the PR to see the reasoning behind this. But I didn’t find anything meaningful. It looks like it was implemented like this from the beginning.

alexander-akait:
@alan-agius4 That is really strange, no one reports about it before :smiley:

alexander-akait:
@alan-agius4 https://github.com/webpack-contrib/css-loader/commit/6da7e90748364fe61231ba71e3d9057c41271f5f, looks like it was implemented for `angular` :smile:

alexander-akait:
Okay, let's merge and release, I think it should not break the logic, because we really have the same in `sourceMappingURL` (maybe chrome had bugs with `sources` before, but I think we can't check)

alan-agius4:
Thanks @alexander-akait

Issue #1540, opened 2023-09-26
getLocalIdent pass selector type


koga73:
### Feature Proposal

I propose that we update the getLocalIdent parameters to pass whether the localName selector is an id or class.

Currently you can pass a custom function for getLocalIdent however there is no way to determine whether the localName parameter is an id or class.

Upon examining the Utils.js file it utilizes "postcss-modules-scope" package and passes a "generateScopedName" function. This function passes (exportName, resourceFile, rawCss) parameters. Currently css-loader only makes use of the exportName parameter. Using the rawCss parameter we can determine if the exportName is an id or class and pass this info to the getLocalIdent function.

### Feature Use Case

I would like the ability to only transform class names and not transform id. This is because keeping the original id is useful for anchoring (href="#section1") as well as code that rely on finding elements by id (automated tests, analytics, vanilla js, etc).

### Please paste the results of `npx webpack-cli info` here, and mention other relevant info

```
  System:
    OS: Windows 10 10.0.19042
    CPU: (32) x64 AMD Ryzen 9 5950X 16-Core Processor
    Memory: 50.14 GB / 63.90 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 116.0.5845.188
    Edge: Spartan (44.19041.423.0), Chromium (117.0.2045.41)
    Internet Explorer: 11.0.19041.1
  Packages:
    babel-loader: 8.2.2 => 8.2.2
    copy-webpack-plugin: 9.0.1 => 9.0.1
    css-loader: 5.0.1 => 5.0.1
    dotenv-webpack: 6.0.0 => 6.0.0
    html-webpack-plugin: 5.5.3 => 5.5.3
    sass-loader: 13.2.2 => 13.2.2
    string-replace-loader: 3.1.0 => 3.1.0
    webpack: 5.88.1 => 5.88.1
    webpack-cli: 5.1.4 => 5.1.4
    webpack-dev-server: 4.15.1 => 4.15.1
```

alexander-akait:
PR welcome

koga73:
PR: https://github.com/webpack-contrib/css-loader/pull/1556
Thank you!

alexander-akait:
[...]

PR #1556, opened 2023-12-21
Updated generateExportEntry to expose node details

koga73:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I would like the ability to only transform class names and not transform id. This is because keeping the original id is useful for anchoring (href="#section1") as well as code that rely on finding elements by id (automated tests, analytics, vanilla js, etc).
https://github.com/webpack-contrib/css-loader/issues/1540

### Breaking Changes

N/A

### Additional Info

Updated postcss-modules-scope dependency with latest which passes the node object to generateExportEntry. The node object has a "type" property that can be used to determine if the node is an id or class.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "postcss": "^8.4.29",
         "postcss-modules-extract-imports": "^3.0.0",
         "postcss-modules-local-by-default": "^4.0.3",
-        "postcss-modules-scope": "^3.0.0",
+        "postcss-modules-scope": "^3.1.0",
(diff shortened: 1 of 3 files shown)


alexander-akait (commented):
Looks good, can you add a small test case to ensure everything works

alexander-akait (approved):

Issue #1621, opened 2024-11-22
incomplete-sanitization in getUrl.js


ruchira-net:
### Bug report

When scanned with CodeQL scanner, it finds a incomplete sanitization issue in the [getUrl.js](https://github.com/webpack-contrib/css-loader/blob/9cf9cd877f934af4aef91f5aaf127467d4964b8e/src/runtime/getUrl.js#L4) file. 

### Actual Behavior

Below method doesn't escape backslash characters in the input.

![image](https://github.com/user-attachments/assets/bccf0399-3840-4a2f-99e8-72cb22d9f1b0)

### Expected Behavior

Method should sanitize untrusted input for preventing injection attacks such as SQL injection or cross-site scripting (Even if the escaped string is not used in a security-critical context, incomplete escaping may still have undesirable effects, such as badly rendered or confusing output).

### How Do We Reproduce?

1. Open the [Wave analysis extension in VSCode](https://marketplace.visualstudio.com/items?itemName=devprod.vulnerability-extension). 
2. Click Add Analysis Tools and select CodeQL.

![image](https://github.com/user-attachments/assets/29b3c9eb-1e0d-4b39-9802-521f05a37370)

4. Click OK

![image](https://github.com/user-attachments/assets/f8e4b3f6-11cb-477d-9277-7ebb9ea68a8a)

5. Run the CodeQL scanner
6. You should see that it complains about the method in the getUrl.js file as below

![image](https://github.com/user-attachments/assets/8fa973a2-ca12-497a-817f-1ef58c9b4630)

### Please paste the results of `npx webpack-cli info` here, and mention other relevant information
  System:
    OS: Windows 11 10.0.26100
    CPU: (12) x64 12th Gen Intel(R) Core(TM) i7-1255U
    Memory: 13.15 GB / 31.69 GB
  Binaries:
    Node: 21.6.2 - C:\Program Files\nodejs\node.EXE
    npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (130.0.2849.46)
    Internet Explorer: 11.0.26100.1882

alexander-akait:
https://github.com/webpack-contrib/css-loader/pull/1622#discussion_r1867595874, It looks like the utility has false positive result here

PR #1622, opened 2024-11-22
#1621 fix: correct escaping of backslashes in URL wrapping

ruchira-net:
This PR contains a:

- [x ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Fixing bug #1621 

### Breaking Changes

None

### DIFF
--- a/src/runtime/getUrl.js
+++ b/src/runtime/getUrl.js
@@ -21,7 +21,7 @@ module.exports = (url, options) => {
   // Should url be wrapped?
   // See https://drafts.csswg.org/css-values-3/#urls
   if (/["'() \t\n]|(%20)/.test(url) || options.needQuotes) {
-    return `"${url.replace(/"/g, '\\"').replace(/\n/g, "\\n")}"`;
+    return `"${url.replace(/\\/g, '\\\\').replace(/"/g, '\\"').replace(/\n/g, "\\n')}"`;
   }
 
   return url;
--- a/src/runtime/getUrl.js
+++ b/src/runtime/getUrl.js
@@ -21,7 +21,7 @@ module.exports = (url, options) => {
   // Should url be wrapped?
   // See https://drafts.csswg.org/css-values-3/#urls
   if (/["'() \t\n]|(%20)/.test(url) || options.needQuotes) {
-    return `"${url.replace(/\\/g, '\\\\').replace(/"/g, '\\"').replace(/\n/g, "\\n')}"`;
+    return `"${url.replace(/\\/g, '\\\\').replace(/"/g, '\\"').replace(/\n/g, "\\n")}"`;
   }
 
   return url;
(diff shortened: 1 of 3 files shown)


ruchira-net:
> Can you add a test case?

Done. Thanks!

test/runtime/getUrl.test.js
```diff
@@ -127,5 +127,6 @@ describe("escape", () => {
         { hash: "#hash", needQuotes: true },
       ),
     ).toMatchSnapshot();
+    expect(getUrl("http://url/path\\/")).toMatchSnapshot();
```
alexander-akait, line 130: Can you provide a real example of the problem? Using `url()`?

ruchira-net: @alexander-akait I've added a more real URL example. Hope that's good enough. Thanks!

src/runtime/getUrl.js
```diff
@@ -21,7 +21,7 @@ module.exports = (url, options) => {
   // Should url be wrapped?
   // See https://drafts.csswg.org/css-values-3/#urls
   if (/["'() \t\n]|(%20)/.test(url) || options.needQuotes) {
```
alexander-akait, line 23: Why we need escape `\\` in runtime? We don't touch such stuff at all, even more, it can literally break already escaped URL

Sorry, we don't need this changes here, even more it worsens the performance

alexander-akait:
https://github.com/webpack-contrib/css-loader/pull/1622#discussion_r1867595874

src/runtime/getUrl.js
ruchira-net: @alexander-akait we are not escaping the \\ in runtime. we are replacing single \ with \\ to escape the URL. \\ is there in the code to escape a single \. Otherwise JavaScript consider single \ as an escaping character itself.

alexander-akait: And what could be the problem? Please provide a **real scenario** that could be used in `url()` and CSS and put it there - https://github.com/webpack-contrib/css-loader/blob/master/test/fixtures/url/url.css, I wouldn't want us to waste perf to something that can never really happen,

ruchira-net: @alexander-akait what if developers have css like this? Does it work right now? I don't think so as the current code doesn't escape the backslashes

`.backslashes {
  background: url("https://www.example.com?path=path\to\resource");
}`

alexander-akait: How does this violate security?

alexander-akait: @ruchira-net Some of them is ident some of them is string, and depending on this we should escaping them or not, I don't see a real integration test of the problem, so I asked to put it in the integration tests and not the unit, I don't want to degrade performance because there is no real problem

PR #1632, opened 2025-04-16
fix: the ASCII whitespaces are preserved so can not be escaped

### LINKED ISSUE #1626: CSS modules with path which contains spaces does not work as expected
### Actual Behavior

generated `src_render_js.js`:

image

source `render.js`:

```js
import * as styles from './space folder/render.module.css';

export function render() {
  const el = document.createElement('div');
  el.className = styles.text; // styles.text contains a space which breaks styling
  document.getElementsByTagName('body')[0].appendChild(el);
  el.innerHTML = 'hello, world';
}
```

### Expected Behavior

https://stackoverflow.com/questions/50812118/is-it-possible-to-use-the-space-character-in-css-class-names

The generated `class` is `src-space folder-render-module__text-WVNeAh` which contains a space, but it's not possible to escape spaces, so the spaces in path should be changed to `__` for example.

### How Do We Reproduce?

https://github.com/JounQin/rspack-css-modules-spaces

Run `pnpm dev:webpack`

JounQin:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

close #1626

### Breaking Changes

N/A

### Additional Info

related https://infra.spec.whatwg.org/#ascii-whitespace

cc @alexander-akait

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -259,8 +259,10 @@ function normalizePath(file) {
   return path.sep === "\\" ? file.replace(/\\/g, "/") : file;
 }
 
+const reversedChartMapper = {};
+
 // eslint-disable-next-line no-control-regex
-const filenameReservedRegex = /[<>:"/\\|?*\s]/g;
+const filenameReservedRegex = /([<>:"/\\|?*\s])/g;
 // eslint-disable-next-line no-control-regex
 const reControlChars = /[\u0000-\u001f\u0080-\u009f]/g;
 
@@ -270,7 +272,21 @@ function escapeLocalIdent(localident) {
     localident
       // For `[hash]` placeholder
       .replace(/^((-?[0-9])|--)/, "_$1")
-      .replace(filenameReservedRegex, "-")
+      .replace(filenameReservedRegex, (_, $1) => {
+        // normalize Windows path `\` as unix `/` for constancy
+        const char = $1 === "\\" ? "/" : $1;
+
+        if (reversedChartMapper[char]) {
+          return reversedChartMapper[char];
+        }
+
+        const hex = char.charCodeAt(0).toString(16).toUpperCase();
+        const escaped = `\\C${hex}`;
+
+        reversedChartMapper[char] = escaped;
+
+        return escaped;
+      })
       .replace(reControlChars, "-")
       .replace(/\./g, "-"),
   );
--- a/src/utils.js
+++ b/src/utils.js
@@ -260,7 +260,7 @@ function normalizePath(file) {
 }
 
 // eslint-disable-next-line no-control-regex
-const filenameReservedRegex = /[<>:"/\\|?*]/g;
+const filenameReservedRegex = /[<>:"/\\|?*\s]/g;
 // eslint-disable-next-line no-control-regex
 const reControlChars = /[\u0000-\u001f\u0080-\u009f]/g;
 
--- a/test/__snapshots__/modules-option.test.js.snap
+++ b/test/__snapshots__/modules-option.test.js.snap
@@ -165,7 +165,7 @@ ___CSS_LOADER_EXPORT___.push([module.id, \`._7-foo-class {
   color: red;
 }
 
-.\\\\--bar-class {
+.\\\\\\\\C3E-bar-class {
   color: red;
 }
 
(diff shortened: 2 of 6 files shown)


test/__snapshots__/modules-option.test.js.snap
```diff
@@ -498,49 +498,49 @@ exports[`"modules" option issue #995 #2: result 1`] = `
...
-.\\1F613 .\\  {
+.\\1F613 .- {
   color: red;
 }
 
-.\\  > .\\  > .\\  {
+.- > .- > .- {
   color: red;
 }
 ",
```
alexander-akait, line 546: Let's add a test case
```css
a-b { }
a\C20b { } /* i.e. with spaces */
```

The main problem we need to generate different CSS selectors for such case.

Even more - I think we should replace each control character on own unique symbol, we can generate it from code the character, so even special characters will always generate and unique selector. What do you think?

alexander-akait: > That could work as expected, but is that common to have different white spaces in a single project?

Usually no, but I think we can solve it more universal for any exotic cases, usually developers don't use such characters in path and in CSS selectors too :smile: 

I think we can just generate use number of a character... yeah it will create more longer CSS selectors, maybe we can use some compression algorithm here to get less length

JounQin: > usually developers don't use such characters in path and in CSS selectors too

I got this issue because we installed a common package in global mode. What means whenever `css-loader` is installed as global, it will be an issue for macOS at least with `Application Support` dir.

JounQin: @alexander-akait Sorry I was busy these days, please help to review again.

src/utils.js
```diff
@@ -270,7 +272,21 @@ function escapeLocalIdent(localident) {
     localident
       // For `[hash]` placeholder
       .replace(/^((-?[0-9])|--)/, "_$1")
-      .replace(filenameReservedRegex, "-")
+      .replace(filenameReservedRegex, (_, $1) => {
+        // normalize Windows path `\` as unix `/` for constancy
```
JounQin, line 276: I'm still not sure how to distinguish `path.sep` vs css `\\` here.

test/__snapshots__/modules-option.test.js.snap
JounQin: Friendly ping @alexander-akait

alexander-akait: @JounQin Sorry for delay too, will review soon

src/utils.js
alexander-akait: hm, I think we can ignore such case, do you see any critical places here?

alexander-akait (commented):
Looks good for, will wait your answer before merge

src/utils.js
JounQin: > hm, I think we can ignore such case, do you see any critical places here?

Nope, the test just fails before current changes here.

alexander-akait (approved):

snitin315 (approved):
LGTM, thanks!

JounQin:
@alexander-akait @snitin315 Any ETA to merge and release?

alexander-akait:
@JounQin this week

JounQin:
Friendly ping @alexander-akait

JounQin:
Friendly ping @alexander-akait @snitin315

JounQin:
Friendly ping @alexander-akait @snitin315  @ersachin3112

JounQin:
@alexander-akait News?

alexander-akait:
@JounQin Oh, sorry, looks like by mistake, but looks like the branch is deleted, I can't reopen, do you try `experiments.css`? It is replace for `css-loader`, `style-loader` and `mini-css-extract-plugin` (faster and without bugs), just want to get feedback, if you need examples feel free to look `webpack/examples/css`, we want to deprecated these packages soon (after the next minor release in favor of `experiments.css`), anyway I can merge it and make a release

JounQin:
@alexander-akait Thanks, that's weird I didn't delete the branch, but can recreate the PR. I'll try `experiments.css` later, but that's a different task, thanks!

JounQin:
@alexander-akait https://github.com/webpack/css-loader/pull/1668

PR #1668, opened 2026-06-08
fix: the ASCII whitespaces are preserved so can not be escaped

### LINKED ISSUE #1626: CSS modules with path which contains spaces does not work as expected

JounQin:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

close #1626

### Breaking Changes

N/A

### Additional Info

related https://infra.spec.whatwg.org/#ascii-whitespace

cc @alexander-akait

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -50,6 +50,7 @@
     "**/fonts/**",
     "node_modules",
     "coverage",
-    "*.log"
+    "*.log",
+    "test/outputs"
   ]
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -260,7 +260,7 @@ function normalizePath(file) {
 }
 
 // eslint-disable-next-line no-control-regex
-const filenameReservedRegex = /[<>:"/\\|?*]/g;
+const filenameReservedRegex = /[<>:"/\\|?*\s]/g;
 // eslint-disable-next-line no-control-regex
 const reControlChars = /[\u0000-\u001f\u0080-\u009f]/g;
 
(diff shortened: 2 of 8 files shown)


JounQin:
cc @alexander-akait

alexander-akait (commented):
Let's fix copilot review and we can merge, note - generation classes with such characters is a little bit different in `experiments.css` than here, but it is fine

JounQin:
@alexander-akait I just fixed the typo and believe the left copilot reviews are invalid, can you confirm?

PR #1679, opened 2026-07-30
fix: strip BOM before process via PostCSS for 8.5.24

### LINKED ISSUE #1678: Leading BOM from a loader-generated module survives into the middle of concatenated CSS (postcss 8.5.24)
### Summary

Since postcss 8.5.24 (["Preserve the BOM when stringifying"](https://github.com/postcss/postcss/pull/2119)), a BOM at the start of a **loader-generated** CSS module survives css-loader's postcss round-trip. When several CSS modules are concatenated into one chunk, that BOM ends up **in the middle** of the file, glued to the next selector — which makes the rule invalid and silently drops it.

dart-sass prepends U+FEFF to its output whenever `style: "compressed"` (the production default) and the output contains non-ASCII; with `expanded` it emits `@charset "UTF-8"` instead. So any Sass module with, say, a `©` in a preserved `/*! … */` banner or a non-ASCII `content:` value produces such a module.

### Minimal reproduction

`src/index.js`

```js
import './first.css';
import './second.scss';
```

`src/first.css`

```css
.first{color:red}
```

`src/second.scss`

```scss
:root{--brand:#0d6efd}
.second{color:var(--brand);content:"©"}
```

`webpack.config.js`

```js
const MiniCssExtractPlugin = require('mini-css-extract-plugin');

module.exports = {
  entry: './src/index.js',
  output: { path: __dirname + '/dist', filename: 'bundle.js', clean: true },
  optimization: { minimize: false },
  module: {
    rules: [
      { test: /\.css$/, use: [MiniCssExtractPlugin.loader, 'css-loader'] },
      {
        test: /\.scss$/,
        use: [
          MiniCssExtractPlugin.loader,
          'css-loader',
          { loader: 'sass-loader', options: { sassOptions: { style: 'compressed' } } },
        ],
      },
    ],
  },
  plugins: [new MiniCssExtractPlugin({ filename: 'styles.css' })],
};
```

### Actual

`dist/styles.css`:

```
.first{color:red}

﻿:root{--brand:#0d6efd}.second{color:var(--brand);content:"©"}
```

In the browser the rule is present but its `selectorText` is `﻿:root`, so it matches nothing: `--brand` resolves to `""` and `.second` falls back to the initial color.

### Expected

No BOM re-emitted for a module that is a **fragment**, not a standalone file — the behaviour up to 8.5.22.

| postcss | BOMs in the emitted chunk |
| --- | --- |
| 8.5.22 | 0 |
| 8.5.24 | 1 |

Reading a `.css` **file** from disk is not affected — css-loader already handles that. Only loader-generated content is.

### Why this matters

In a real Bootstrap 5 project this dropped the entire `:root,[data-bs-theme=light]` block — ~550 custom properties including fonts, colors and `--bs-border-*` — so the whole site rendered unstyled. The same stray BOM additionally crashed cssnano's `postcss-merge-rules` when it landed in front of a `/*!` banner containing a comma.

### Suggested fix

Strip a leading BOM from the module source before handing it to postcss (or from the stringified result). css-loader's output is a fragment; postcss's change is reasonable for the single-file round-trip, and the fragment case is exactly where it breaks.

### Workaround

`sassOptions: { charset: false }` in sass-loader.

### Versions

css-loader 7.1.4 · postcss 8.5.24 · sass 1.94.3 · sass-loader 16.0.8 · mini-css-extract-plugin 2.9.x · webpack 5.109.1 · Node 20

ShGKme:
**Summary**

- Fix: https://github.com/webpack/css-loader/issues/1678
- Ref: https://github.com/postcss/postcss/pull/2119

1. Some loaders during a single CSS file processing add BOM
   - For example, `sass-loader` with the default production configuration:
     `{ sassOptions: { outputStyle: 'compressed', charset: true }}`
   - The result is valid as a single file compilation result
2. `css-loader` processes CSS via `postcss`
   - Before `postcss@8.5.24` it striped BOM
   - After `postcss@8.5.24` it preserves BOM
4. Problems:
   1. The result is different depending on the installed PostCSS patch version
   2. The result cannot be concatenated (for example, with `mini-css-extract-plugin`) - BOM will be in the middle.

This PR strips BOM before passing to `postcss`, basically copying `postcss`'s behavior before  `postcss@8.5.24`.

Alternatives could be:
- Strip BOM after processing, not before
- Consider it a problem of the concatenation side

Yes, but I'm unsure if this is the kind of tests that are expected.
To have a propper integration test, we need to install postcss 8.5.24 or later.
Please let me know if I need to adjust the tests.

No. It results is the same output from both `postcss` before and after `v8.5.24`.

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

I don't think this needs to be documented because.

**Use of AI**

- No AI was used

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -28,6 +28,7 @@ import {
   warningFactory,
   syntaxErrorFactory,
   supportTemplateLiteral,
+  stripBom,
 } from "./utils";
 
 export default async function loader(content, map, meta) {
@@ -174,7 +175,8 @@ export default async function loader(content, map, meta) {
   let result;
 
   try {
-    result = await postcss(plugins).process(content, {
+    // Since PostCSS 8.5.24 it does not strip BOM anymore
+    result = await postcss(plugins).process(stripBom(content), {
       hideNothingWarning: true,
       from: resourcePath,
       to: resourcePath,
--- a/src/utils.js
+++ b/src/utils.js
@@ -1471,6 +1471,11 @@ function supportTemplateLiteral(loaderContext) {
   return false;
 }
 
+function stripBom(content) {
+  const BOM = 0xfeff;
+  return content.charCodeAt(0) === BOM ? content.slice(1) : content;
+}
+
 export {
   normalizeOptions,
   shouldUseModulesPlugins,
@@ -1499,4 +1504,5 @@ export {
   warningFactory,
   syntaxErrorFactory,
   supportTemplateLiteral,
+  stripBom,
 };
--- a/test/__snapshots__/loader.test.js.snap
+++ b/test/__snapshots__/loader.test.js.snap
@@ -63,6 +63,42 @@ exports[`loader should not generate console.warn when plugins disabled and hideN
 
 exports[`loader should not generate console.warn when plugins disabled and hideNothingWarning is "true": warnings 1`] = `[]`;
 
+exports[`loader should not path BOM to postcss (issue #1678): errors 1`] = `[]`;
+
+exports[`loader should not path BOM to postcss (issue #1678): module 1`] = `
+"// Imports
+import ___CSS_LOADER_API_NO_SOURCEMAP_IMPORT___ from "../../../../src/runtime/noSourceMaps.js";
+import ___CSS_LOADER_API_IMPORT___ from "../../../../src/runtime/api.js";
+var ___CSS_LOADER_EXPORT___ = ___CSS_LOADER_API_IMPORT___(___CSS_LOADER_API_NO_SOURCEMAP_IMPORT___);
+// Module
+___CSS_LOADER_EXPORT___.push([module.id, \`.first::after {content:"©";}
+\`, ""]);
+// Exports
+export default ___CSS_LOADER_EXPORT___;
+"
+`;
+
+exports[`loader should not path BOM to postcss (issue #1678): module 2`] = `
+"// Imports
+import ___CSS_LOADER_API_NO_SOURCEMAP_IMPORT___ from "../../../../src/runtime/noSourceMaps.js";
+import ___CSS_LOADER_API_IMPORT___ from "../../../../src/runtime/api.js";
+var ___CSS_LOADER_EXPORT___ = ___CSS_LOADER_API_IMPORT___(___CSS_LOADER_API_NO_SOURCEMAP_IMPORT___);
+// Module
+___CSS_LOADER_EXPORT___.push([module.id, \`.second::after{content:"©";}
+\`, ""]);
+// Exports
+export default ___CSS_LOADER_EXPORT___;
+"
+`;
+
+exports[`loader should not path BOM to postcss (issue #1678): result 1`] = `
+".first::after {content:"©";}
+.second::after{content:"©";}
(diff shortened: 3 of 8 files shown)


vladopol:
Supporting this PR with data from a released application.

Nextcloud Talk Desktop 2.2.2 shipped with **8 rules silently dropped across 4 CSS chunks** — the reply-quote layout, file previews, message body, shared items list, and `._material-design-icon` in three chunks. Reply quotes rendered as full-size blurred images instead of thumbnails, identically on macOS, Windows and Linux. The project tracked it as nextcloud/talk-desktop#1836 and released 2.2.3 the next day, pinning `postcss@8.5.23` and adding a dependabot `ignore` for `^8.5.24` (nextcloud/talk-desktop#1837). We are downstream of that project — we run it and build it from source in-house — and the numbers below come from our own builds.

Worth spelling out: that application's webpack config passes a bare `'sass-loader'` with no options. `sass-loader` sets `style: "compressed"` by itself in production mode, so no explicit `sassOptions` is needed to trigger this — one non-ASCII byte anywhere in the Sass sources is enough.

The failure is silent, which is what makes it expensive:

| Check | Result |
| --- | --- |
| the rule is present in the emitted `.css` | yes, fully intact |
| the chunk is loaded (`performance.getEntriesByType('resource')`) | yes |
| the rule is present in `document.styleSheets` | yes |
| `el.matches('.quote[data-v-7398ef7c]')` | `true` |
| `getComputedStyle(el).display` | `block` — the rule never applies |

Nothing in the build warns. It is found by a person noticing that a component looks wrong, after release.

This is also not a new design question here. In webpack/mini-css-extract-plugin#821 (2021) @alexander-akait described this exact failure mode — *"merging modules in one file using UTF-8 BOM is not safe because you will have BOM in the middle of the file"* — and named the remedy: *"Ideally `css-loader` should trim `@charset` or `UTF-8 BOM` and keep it as property of modules"*. That issue was closed with BOM explicitly unsupported in webpack, and webpack core has since applied the same rule to its own module types (`lib/util/removeBOM.js`, applied per module through `NormalModule.processResult` in `JavascriptModulesPlugin` and `CssModulesPlugin` — webpack#19275). This PR applies it to the one path still missing it. Trimming is also the half that stops the corruption on its own — keeping the charset as a property of the module, as that comment went on to suggest, is separable and can follow in its own change (css-loader#1212 covers the `@charset` side of it).

Waiting for postcss to reconsider will not help: in postcss#2122 the maintainer confirmed that preserving the BOM is intended and that downstream tools should handle it.

The two workarounds available today are not equivalent, and both are application-level:

| postcss | `sassOptions.charset: false` | BOMs in emitted CSS |
| --- | --- | --- |
| 8.5.23 | no | 0 |
| 8.5.25 | no | 8 |
| 8.5.25 | yes | 0 |

(the third row is the control: without it, the second only proves the postcss swap took effect)

A pin holds until someone unpins it, and `charset: false` has to be repeated in every loader rule of every application. This PR is the only variant that does not.

alexander-akait:
Fixed in the main, sorry for delay, release will be soon

